PR #9, opened 2026-08-21
Land the SmartHome project on main: cleanup, Homie v4 fixes, on-device test suite

JojoEffect:
Brings the project onto `main`, which currently holds three files (`.gitignore`, `LICENSE`, `README.md`). Everything else — `src/`, `scripts/`, `tools/`, the docs — has only ever existed on branches, so the diff here is effectively the whole repository plus 29 commits of work on top.

## Why it matters beyond the code

Until this lands, a fresh clone gets an empty repository, and the `@claude` GitHub workflow cannot fire at all: GitHub reads workflow files for `issue_comment` events from the **default branch**, and `.github/workflows/claude.yml` is not on it.

## What is in it

**Repository structure and tooling**
- Test projects split by kind: `src/devices` (real apps), `src/integrationTests` (on-device end-to-end checks), `src/tests` (unit tests).
- One naming rule: assembly = root namespace = `SmartHome.<Area>.<Name>`, folders keep the short name. The unit test assembly is the single documented exception — nanoFramework's `UnitTestLauncher` resolves it by the name `NFUnitTest`.
- The reconnecting MQTT client moved out of the Homie library into `SmartHome.Mqtt`. It never contained anything Homie-specific; the Homie publishing lives in extension methods that hang off M2Mqtt's own interface.
- Copilot instructions and the vendored multi-tool "caveman" skill packs removed (7 directories, 152 files); `CLAUDE.md` is the single source of truth.
- Dev scripts consolidated: one entry point per workflow, `Stop-DevEnv.ps1` added, `Start-AgentWorkspace.ps1` folded into `Start-DevEnv.ps1`.

**Homie v4 conformance fixes**
- The last will was never declared. The device app connected the transport first, so `HomieClient` skipped its own connect and silently discarded the will, keepalive and credentials — a device that died showed `$state: ready` forever. The client now owns its session.
- Settable properties were subscribed on the property *value* topic instead of `.../set`, so controller commands were never received and the device re-consumed its own retained publishes.
- `$extensions` is mandatory and was never published.
- The device now re-announces after a reconnect. Homie state lives in the broker's retained store, so a restarted broker had never heard of the device.
- Topic ids are validated at construction against the spec's rules.
- `IHomieClient` extracted, with `Alert()`/`Sleep()`/`Ready()` (previously unreachable — `TryChangeState` is internal) and an `OnCommand` event that distinguishes a controller's `/set` from the device's own update.

**Tests**
- Unit suite 25 tests. `Run-Tests.ps1` now decides from TRX counters: vstest exits 0 when every test is *skipped*, which hid a broken run for three commits.
- Integration suite of five on-device checks with a single entry point (`Run-IntegrationTests.ps1`), covering WiFi, an MQTT round trip, the BMP280 sensor, broker-outage recovery, and a device-agnostic Homie conformance check that asserts attributes, retained flags, every datatype, `/set` round-trips, the `alert`/`sleeping` states and re-announcement.

**RoomSensor**
- Publishes real BMP280 readings; it previously filled temperature, humidity and pressure with `random.NextDouble()`.
- Declares `$unit` for all three properties, and moves to `alert` when a reading is invalid.

## Verification

Everything was run on the physical ESP32 (COM3), not just built:

- Integration suite 5/5, unit suite 25/25 executed and passed.
- The last will confirmed on the broker (`Will message specified (4 bytes) (r1, q1)`) and delivered on an unclean drop.
- Re-announcement confirmed by destroying the broker under a running device and watching the full announcement reappear.
- RoomSensor's readings cross-checked against the isolated sensor test.

Suite runtime is 240s, down from 459s: captures now stop when the device reports instead of sitting out a fixed window.

## Known open items

Tracked in `NEXT-STEPS.md`: float payloads don't round-trip their own value (nanoFramework's `double.ToString()`), and `IrrigationControl`/`OvenControl` are still empty stubs pending a decision on what hardware they drive.


src/common/Homie/V4/HomieClient.cs
```diff
@@ -0,0 +1,373 @@
...
+        /// is exactly what this code used to do. A foreign session is therefore
+        /// replaced, not reused.
+        /// </remarks>
+        public bool Connect()
+        {
+            _logger.LogDebug("Connect...");
+
+            try
+            {
+                _device.OnDeviceStateChange += HandleDeviceStateChange;
```
JojoEffect, line 127: **Handler leak: `Connect()` subscribes without ever unsubscribing.**

`_device.OnDeviceStateChange += HandleDeviceStateChange;` runs on every `Connect()` call, and nothing removes it. Both device apps retry (`RoomSensor/Program.cs:156`, `HomieClientCheck/Program.cs:151`), and this `+=` happens *before* the connect attempt, so a first attempt that fails leaves a handler attached.

On the second, successful attempt `TryChangeState(State.Init)` fires the event twice:
- handler A publishes device info and moves the device to `ready`;
- handler B, still in the `Init` invocation, publishes device info again and calls `TryChangeState(State.Ready)` from an already-`ready` state, which `CanChangeState` rejects â€” taking the failure branch at line 207 and calling `DisconnectInternal()`.

`DisconnectInternal()` calls `_mqttClient.Disconnect()`, which sets `_autoReconnectEnabled = false`. So `Connect()` returns **true** while the device is disconnected with auto-reconnect switched off, and stays dark until it is power-cycled.

The trigger is precisely the case the retry loop exists for â€” a broker that isn't up yet, which `Run-IntegrationTests.ps1` induces deliberately by cycling the broker. Every unit test calls `Connect()` exactly once, which is why it is invisible today.

`RegisterPropertyUpdateHandlers` (line 313) already guards this with a `-=` before its `+=`; the same is needed here.

```suggestion
                _device.OnDeviceStateChange -= HandleDeviceStateChange;
                _device.OnDeviceStateChange += HandleDeviceStateChange;
```

src/common/Homie/V4/HomieClient.cs
```diff
@@ -0,0 +1,373 @@
...
+            _settablePropertiesTable.Keys.CopyTo(topics, 0);
+
+            var qosLevels = new MqttQoSLevel[topics.Length];
+            for (int i = 0; i < qosLevels.Length; i++)
+            {
+                qosLevels[i] = MqttQoSLevel.AtLeastOnce;
+            }
+
+            _mqttClient.Subscribe(topics, qosLevels);
+            _mqttClient.MqttMsgPublishReceived += HandleIncomingMessage;
```
JojoEffect, line 333: **Same handler leak, different symptom: every `/set` command fires N times.**

`SubscribeSettablePropertyTopics` attaches `HandleIncomingMessage` with no matching `-=`, and it runs on every `Connect()`. After one failed and one successful attempt the handler is attached twice, so a single controller publish to `.../set` applies `property.Set()` twice, republishes the value twice, and raises `OnCommand` twice.

`OnCommand` exists precisely for the actuators in `NEXT-STEPS.md` item 3 â€” one command opening a valve or starting a heater twice is the failure this becomes.

```suggestion
            _mqttClient.MqttMsgPublishReceived -= HandleIncomingMessage;
            _mqttClient.MqttMsgPublishReceived += HandleIncomingMessage;
```

src/common/Homie/V4/HomieClient.cs
```diff
@@ -0,0 +1,373 @@
...
+            _logger.LogInformation("MQTT connection closed handler called.");
+
+            UnregisterPropertyUpdateHandlers();
+        }
+
+        private void RegisterConnectionChangeHandlers()
+        {
+            _logger.LogDebug("Registering connection change handlers...");
+
+            _mqttClient.ConnectionClosed += HandleConnectionClosed;
```
JojoEffect, line 477: **Same handler leak on the connection-change handlers.**

`ConnectionClosed` and `ConnectionOpened` are attached with `+=` on every `Connect()` and never removed. With two attachments a single broker reconnect runs `HandleConnectionOpen` twice, so the whole `init -> ready` re-announce happens twice â€” every device, node and property attribute published twice per reconnect.

```suggestion
            _mqttClient.ConnectionClosed -= HandleConnectionClosed;
            _mqttClient.ConnectionOpened -= HandleConnectionOpen;
            _mqttClient.ConnectionClosed += HandleConnectionClosed;
            _mqttClient.ConnectionOpened += HandleConnectionOpen;
```

src/common/Mqtt/ReconnectingMqttClient.cs
```diff
@@ -0,0 +1,273 @@
...
+
+        public void Disconnect()
+        {
+            _logger.LogDebug("Disconnect from MQTT broker...");
+            _autoReconnectEnabled = false;
+
+            _mqttCient.ConnectionClosed -= ReconnectHandler;
+            _mqttCient.ConnectionClosedRequest -= ReconnectHandler;
+
+            _connectThread?.Abort();
```
JojoEffect: **Reconnect thread is aborted outside the lock that guards it.**

`ReconnectHandler` creates and assigns `_connectThread` inside `lock (_reconnectSync)`, but `Disconnect()` reads and aborts it outside that lock, on a different thread (M2Mqtt raises `ConnectionClosed` from its receive thread).

If `Disconnect()` runs between the handler's `_autoReconnectEnabled` check and its assignment, it sees `null`, aborts nothing, and a reconnect thread starts after `Disconnect()` has completed. The loop condition stops it, so the usual outcome is benign â€” but the `Abort()` itself is the real risk: landing inside `ConnectInternal` can leave the socket half-open on nanoFramework, and `Disconnect()` has already cleared `_subscribedTopics`, so a connection that did get established would come back with no subscriptions.

Taking the lock around the abort, and checking `_autoReconnectEnabled` once inside it, closes the window.

scripts/Run-IntegrationTests.ps1
```diff
@@ -0,0 +1,774 @@
...
+        [string]$Port
+    )
+
+    $deviceId = $Settings.DeviceId
+    $nodeId = $Settings.NodeId
+    $root = "homie/$deviceId"
+    $node = "$root/$nodeId"
+    $failures = @()
+
+    function Add-Failure {
```
JojoEffect: **Dead helper that the next assertion will use by mistake.**

`Add-Failure` is defined here and never called â€” every assertion site appends to `$script:conformanceFailures` directly. It reads as the intended way to record a failure, so the next person adding an assertion will call it, and their failure will land in a collection the verdict at line 486 never reads: a silently passing conformance test.

Either delete it, or route every site through it.

scripts/Run-IntegrationTests.ps1
```diff
@@ -0,0 +1,774 @@
...
+        [string]$Topic,
+        [string]$Expected,
+        [int]$TimeoutSeconds
+    )
+
+    $deadline = (Get-Date).AddSeconds($TimeoutSeconds)
+    $seen = '<nothing>'
+
+    while ((Get-Date) -lt $deadline) {
+        $snapshot = Get-HomieRetainedSnapshot -Port $Port -SettleSeconds 2
```
JojoEffect, line 561: **Each poll iteration spawns a subscriber and sleeps 2s.**

`Wait-ForRetainedValue` calls `Get-HomieRetainedSnapshot` in its loop, and every call starts a fresh `mosquitto_sub`, waits 2s and kills it. `HomieClientCheck` goes through this at least eleven times (settle, five `/set` reflections, four lifecycle states, re-announce) plus two 3s snapshots, so roughly 30s of the test's 108s is spent waiting out snapshot windows even when the value had already arrived â€” and this test is 45% of a 240s suite.

The fresh subscriber is genuinely required (a retained message only carries the retain flag when replayed to a *new* subscriber), but one subscriber can serve a batch: publish all five `/set` commands, then take a single snapshot and check all five reflections against it.

scripts/Deploy-ToDevice.ps1
```diff
@@ -0,0 +1,168 @@
...
+$padded[0] = 0xFF
+$filled = 1
+while ($filled -lt $PaddedImageSize) {
+    $chunk = [Math]::Min($filled, $PaddedImageSize - $filled)
+    [Array]::Copy($padded, 0, $padded, $filled, $chunk)
+    $filled += $chunk
+}
+
+[Array]::Copy($imageBytes, $padded, $imageBytes.Length)
+[System.IO.File]::WriteAllBytes($paddedImage, $padded)
```
JojoEffect, line 154: **The padded image is written on every deploy and never cleaned up.**

`<assembly>.padded.bin` is `$PaddedImageSize` bytes (400KB) written into `bin\Debug` and left there â€” `nanoff` has already consumed it by the time the script exits. Five flashes per suite run across six deployable projects means megabytes of dead build output accumulating in the tree.

It also leaves a stale `.padded.bin` sitting next to a freshly built `.bin`, which is the same shape as the stale-assembly problem this repo documents in `Deploy-ToDevice.ps1`'s own comments. Deleting it after `nanoff` returns, or writing it under `$env:TEMP`, avoids both.

PR #30, opened 2026-08-22
fix(homie): publish float values a controller can read back

### LINKED ISSUE #10: Float property values do not round-trip their own payload
`FloatProperty` renders its payload with `double.ToString()`. nanoFramework's
implementation renders `21.5` as `21.499999999999999`, so the value a controller reads
back is not the value the device meant to publish, and the device cannot parse its own
retained payload back to the same double after a reconnect.

The conformance check does not catch this: its `/set` verification compares parsed
doubles rather than payload strings, precisely so this behaviour would not fail the
suite.

Needs a formatting decision before any code changes. Check what
`nanoFramework.CoreLibrary` actually supports first — the pinned sibling checkout at
`../CoreLibrary` is the reference, and round-trip (`"R"`) formatting may not be
available. Candidates: a fixed number of decimal places per property, a `$format`-aware
precision, or a hand-rolled formatter.

Whatever is chosen has to hold for `IntegerProperty` too, and the host-side comparison
in `Run-IntegrationTests.ps1` should then be tightened to compare strings, so the suite
starts guarding it.

**Files:** `src/common/Homie/V4/Properties/FloatProperty.cs`

JojoEffect:
## Summary

A float property set to `21.5` published **`21.499999999999999`**. That is nanoFramework's `double.ToString()`, which uses `"G"`. Closes #10.

It is **value-dependent** — `0.1` rendered correctly — which is how it survived long enough to reach a shipped device. RoomSensor publishes three float properties, so every temperature, humidity and pressure reading carried it.

## The formatting decision, measured rather than reasoned about

The issue asked for a decision. I probed the actual formatter on the virtual device:

| value | `ToString()` (`G`) | `"R"` | `"F2"` | `"N2"` |
|---|---|---|---|---|
| 21.5 | **21.499999999999999** | **throws** | 21.50 | 21.50 |
| 1234.5678 | 1234.5678 | **throws** | 1234.57 | **1,234.57** |
| 0.1 | 0.1 | **throws** | 0.10 | 0.10 |
| 1e-7 | 1E-07 | **throws** | 0.00 | 0.00 |

Round-trip format `"R"` — the obvious answer — throws `NotImplementedException` on this runtime. `"N"` inserts a **thousands separator** that no controller could parse. `"F<n>"` was correct across everything tested.

So float properties now render at a declared precision: **two decimals by default**, set per property with `WithDecimals`. Two is sensor-shaped — natural for temperature, humidity and pressure, and parses back exactly. A percentage rarely needs two places and a pressure in Pascals might want none, which is what the builder method is for.

## One defence for a latent problem

Homie requires a dot. nanoFramework's `Double` exposes **no `ToString(format, IFormatProvider)` overload**, so the separator cannot be pinned at the call site — the formatter reads `NumberFormatInfo.CurrentInfo`, which is invariant today *only because* nothing references `nanoFramework.System.Globalization`. Adding that package anywhere in the solution would silently start publishing `21,50`.

The formatter repairs a comma character-wise, because nanoFramework's `String` has no `Replace` at all.

## The test that was hiding it

The conformance check compared float echoes **numerically with a 0.001 tolerance**, with a comment explaining it was there to accommodate `21.4999...`. That tolerance was tolerating the defect rather than measuring it: what a controller reads is a *string*, and `21.499999999999999` is wrong whatever it parses to.

It now compares exact strings, with the expected echo spelled out where rendering differs from what was sent — a `/set` of `21.5` must come back as `21.50` from a two-decimal property. That is the assertion that guards this fix.

## Testing Done

- **38/38 unit tests** on the nanoclr virtual device, four of them new: the round-trip payload, the decimal separator, per-property precision (0 and 4 places), and rejection of an impossible precision.
- **No hardware run yet.** The conformance-check change and RoomSensor's altered output both need a device — see below.

## Checklist

- [x] Relevant tests pass
- [x] PR title follows Conventional Commits

## Before merging

This changes what a shipped device publishes: RoomSensor's readings go from `22.798262261926721` to `22.80`. The unit tests cover the formatting, but the conformance check's new exact-string comparison has only ever run against the *old* behaviour. It wants one integration run on the ESP32 before this is trusted.

Note the second commit restores `Unit.nfproj`: a `sed -i` used while probing the formatter rewrote it from CRLF to LF. The probe was removed and there is no intended change to that file, so the net branch diff excludes it entirely.


JojoEffect:
## Hardware verification — done

Supersedes the "no hardware run yet" note in the description.

**Integration suite 5/5 on the ESP32 (COM3), 195s.**

The one that matters is `HomieClientCheck`, which now compares float echoes as **exact strings**: it publishes `21.5` to the float property's `/set` topic and requires exactly `21.50` back. It passed, so the device genuinely renders at the declared precision rather than merely landing within a tolerance.

```
  WifiCheck            PASS   30s
  MqttCheck            PASS   20s
  Bmp280Check          PASS   27s
  HomieClientCheck     PASS   64s  attributes, datatypes, retained flags, /set round-trip, alert/sleeping and re-announce all conform
  MqttReconnectCheck   PASS   54s
  TOTAL                      195s
```

**RoomSensor redeployed**, and what it publishes has visibly changed:

| | before | after |
|---|---|---|
| temperature | `23.014640818774934` | `23.01` |
| humidity | `51.817708124258441` | `51.85` |
| pressure | `94612.663…` | `94612.66` |

The initial announce carries `0.00` rather than `0`, which is the same formatter applied to the initial value — correct and consistent.

One thing worth not misreading: `Bmp280Check` still reports `23.014640818774934°C` in its summary line. That is the test's own `Debug.WriteLine` of the raw sensor reading, not a Homie payload — it has no Homie properties at all, so it is unaffected by this change and correctly so.

JojoEffect:
## Note on the hardware run above

The 5/5 integration result was recorded against `5c0e1a6`, **before** the review commits. Being precise about what that does and does not still cover:

**Still covered.** The formatting path is untouched by the review fixes, so the exact-string conformance assertion — a `/set` of `21.5` coming back as `21.50` — still stands on that evidence.

**Not covered.** The review added a guard rejecting NaN and the infinities. Neither the suite nor RoomSensor can produce a non-finite value (RoomSensor checks the sensor's `IsValid` flags first), so the suite would not exercise the new path either way. It is covered by unit tests instead, on the virtual device.

The one thing a re-run would add is confirmation that the guard did not disturb the happy path — that RoomSensor still publishes normally with an extra check on every `Update`. Low risk, but it is a hardware claim I would rather not carry over silently.

JojoEffect:
## Re-run against the review commits — clean

Replaces the caveat in my previous comment. Everything below is against `HEAD` of this branch, not the earlier commit.

**Integration suite 5/5 on the ESP32 (COM3), 206s.**

```
  WifiCheck            PASS   30s
  MqttCheck            PASS   32s
  Bmp280Check          PASS   26s
  HomieClientCheck     PASS   62s  attributes, datatypes, retained flags, /set round-trip, alert/sleeping and re-announce all conform
  MqttReconnectCheck   PASS   56s
  TOTAL                      206s
```

`HomieClientCheck` passes with the exact-string float comparison intact — a `/set` of `21.5` still comes back as exactly `21.50`.

**RoomSensor redeployed and observed for 50s**, which is what the re-run was actually for: confirming the non-finite guard now running on every `Update` does not disturb the happy path.

```
homie/room-sensor-office/sensor/temperature -> '22.90'
homie/room-sensor-office/sensor/humidity    -> '52.07'
homie/room-sensor-office/sensor/pressure    -> '94749.25'
homie/room-sensor-office/sensor/temperature -> '22.91'
homie/room-sensor-office/sensor/humidity    -> '52.06'
homie/room-sensor-office/sensor/pressure    -> '94748.42'
```

Steady loop, no exceptions, no `alert` transitions.

To be exact about what is and is not proven: the guard's *rejection* path is covered by unit tests on the virtual device, not here — neither the suite nor RoomSensor can produce a non-finite value, which is the point. What this run establishes is that adding the check costs nothing on the path that actually runs.

Issue #16, opened 2026-08-22
Retained-store polling uses fixed sleeps rather than bounded polls


JojoEffect:
`Get-HomieRetainedSnapshot` starts a subscriber, sleeps a fixed 2–3 seconds, then kills
it. On localhost the broker replays its whole retained store within milliseconds of
SUBACK, so nearly all of that is idle time — and `Wait-ForRetainedValue` pays it again
on every poll iteration. Roughly 15–25s of `HomieClientCheck`'s 63s.

Cheaper, without losing the semantics:

- Bound the poll on the real condition — return the moment the wanted line appears in
  the output file, polling at 50–100ms.
- For a whole-store read, stop once the file has gone ~150ms without growing; the
  retained replay is one contiguous burst.
- For single-topic waits, `mosquitto_sub -t <topic> -C 1 -W 1` self-terminates on the
  first retained message, which also removes the `cmd.exe` wrapper and the `taskkill`.

**Must keep:** the subscriber has to be *fresh*. MQTT only sets the retain flag when
replaying from the store to a new subscriber, so the long-running `homie/#` log cannot
answer "is this retained".

**Files:** `scripts/Run-IntegrationTests.ps1`

JojoEffect:
## Investigated, and the answer is no — with evidence

Worked in #31. The premise here ("the fixed sleeps are the cost, bound the poll instead") does not survive measurement, and I would rather record why than leave this open for someone to retry.

**The sleeps were not the dominant cost.** Instrumenting a snapshot:

```
Start-Process returned after : 0.017s
first byte in the file       : 2.629s
```

Nothing arrived for 2.6 seconds. The cause was `localhost` resolving to `::1` first against an IPv4-only broker — 2.03s versus 0.02s for `127.0.0.1`. That part is fixed in #31 and is a real improvement, though mostly to *reliability*: a 2s window had been almost entirely connect time, leaving near-zero listening.

**Bounding the poll does not work.** Three attempts, three hardware failures:

| attempt | result |
|---|---|
| stop after 250ms of quiet output | 7 conformance failures |
| flat 1s window | 3 failures, 54 topics where a healthy run sees ~64 |
| `mosquitto_sub -W 1`, wait for exit | complete, but **6.30s** — twice as slow as the 3s it replaces |

An A/B against known ground truth (80 seeded retained topics) settled it:

```
OLD: flat 3s sleep, then kill    3.31s   80/80   complete
[...]

PR #31, opened 2026-08-23
fix(tests): require the retain flag, and stop paying an IPv6 timeout

JojoEffect:
## Summary

This started as #16 ("replace the fixed sleeps, save ~20s") and turned into something else. It does **not** make the suite faster — it makes it about 8s slower — and it fixes a latent bug in the conformance check that had been masked by a two-second delay.

Related to #16; deliberately does **not** close it. See the comment there.

## Two real changes

**1. `localhost` cost two seconds on every mosquitto client call.**

| host | time to first message |
|---|---|
| `localhost` | 2.03s, 2.03s, 2.06s |
| `127.0.0.1` | 0.02s, 0.02s, 0.12s |

`localhost` resolves to `::1` first while `Start-DevEnv.ps1` binds `0.0.0.0` — IPv4 only, deliberately, so a real device on the LAN can reach it. Every client attempted IPv6, waited out the connect timeout, then fell back.

This is not primarily a speed fix. A 2s snapshot window used to be consumed almost entirely by the connect, leaving close to **no listening time**; the polling loop only hid that by retrying until something turned up. The same window is now actual listening.

**2. `Wait-ForRetainedValue` ignored the retain flag** — and that is what the whole check exists to verify.

It accepted any snapshot whose topic held the expected payload. A snapshot subscriber connecting while the device is still announcing receives the rest of that announce **live**, and a live delivery carries `retain=0`. Such a snapshot was accepted and handed to the caller, which then reported a conforming device's attributes as not retained.

It never fired before because the 2s IPv6 timeout reliably delayed the subscriber past the end of the announce. **The conformance check was passing on an accident of timing.** Removing the delay produced 31 spurious failures, which is how it surfaced.

Requiring the flag also makes reusing the snapshot sound rather than merely plausible: `$state` is published last, so a subscriber that saw it *replayed* necessarily connected after everything before it was in the store. That was already the stated justification; it is now enforced.

## What I got wrong on the way

Three hardware runs were spent blaming the settle window. The commits are left in history because the measurements are the useful part:

- **Quiet-detection (250ms of no output)** — 7 failures. I asserted the replay is "one contiguous burst". It is not.
- **Flat 1s window** — 3 failures, 54 topics where a healthy run sees ~64.
- **Restoring 3s** — 31 failures. This is what proved the window was never the variable.

An A/B against known ground truth (80 seeded topics) settled it:

```
OLD: flat 3s sleep, then kill    3.31s   80/80   complete
OLD: flat 1s sleep, then kill    1.28s   80/80   complete
NEW: -W 1, wait for exit         6.30s   80/80   complete
```

The kill loses nothing against a static store, so buffering was never the cause, and `mosquitto_sub -W` is twice as slow as what it would replace. My first two explanations were both wrong.

**The earlier harness missed all of this** because it compared the new implementation's topic count against *itself* across four runs. A systematically truncated snapshot passes that perfectly. Ground truth is the whole point.

## Testing Done

**Integration suite 5/5 on the ESP32, 214s.** `HomieClientCheck` passes, including the retained-flag assertions that were failing.

Cost, stated plainly:

| | baseline | this branch |
|---|---|---|
| HomieClientCheck | 62–64s | 74s |
| suite total | 195–206s | 214s |

The extra poll needed to obtain a genuinely-replayed snapshot is the price of the check actually checking. Worth it.

## Checklist

- [x] Relevant tests pass
- [x] PR title follows Conventional Commits

### DIFF
--- a/scripts/Common.ps1
+++ b/scripts/Common.ps1
@@ -333,7 +333,18 @@ function Get-SmartHomeSubscriberArguments {
     # ("All messages MUST be sent as retained, UNLESS stated otherwise") and only the
     # broker can confirm it, but -v prints topic and payload only. Lines become
     # "<topic> <0|1> <payload>", which readers still match by topic prefix.
-    return @('-h', 'localhost', '-p', $Port, '-t', $SmartHomeHomieTopic, '-F', '%t %r %p')
+    # 127.0.0.1, not 'localhost'. Measured, not guessed: 'localhost' resolves to ::1
+    # first, and Start-DevEnv.ps1 binds the broker to 0.0.0.0 -- IPv4 only, deliberately,
+    # so a real device on the LAN can reach it. Every client therefore attempts IPv6,
+    # waits out the connect timeout, and only then falls back.
+    #
+    #   mosquitto_sub -h localhost   first message after 2.03s, 2.03s, 2.06s
+    #   mosquitto_sub -h 127.0.0.1   first message after 0.02s, 0.02s, 0.12s
+    #
+    # Two seconds on every subscriber and every publish. The integration suite takes
+    # about a dozen retained-store snapshots per conformance run, so this alone was
+    # roughly half that test's duration.
+    return @('-h', '127.0.0.1', '-p', $Port, '-t', $SmartHomeHomieTopic, '-F', '%t %r %p')
 }
 
 # ── Dev environment: state ────────────────────────────────────────────────────
--- a/scripts/Common.ps1
+++ b/scripts/Common.ps1
@@ -298,6 +298,25 @@ function Invoke-GitCloneOrUpdate {
 $SmartHomeDevEnvPrefix = 'smarthome'
 $SmartHomeHomieTopic = 'homie/#'
 
+# The address host-side mosquitto clients dial. 127.0.0.1, not 'localhost'. Measured, not
+# guessed: 'localhost' resolves to ::1 first, and Start-DevEnv.ps1 binds the broker to
+# 0.0.0.0 -- IPv4 only, deliberately, so a real device on the LAN can reach it. Every
+# client therefore attempts IPv6, waits out the connect timeout, and only then falls back.
+#
+#   mosquitto_sub -h localhost   first message after 2.03s, 2.03s, 2.06s
+#   mosquitto_sub -h 127.0.0.1   first message after 0.02s, 0.02s, 0.12s
+#
+# Two seconds on every subscriber and every publish, and worse than slow: a 2s snapshot
+# window used to be spent almost entirely connecting, so it captured nearly nothing and
+# the polling loops only hid that by retrying.
+#
+# Deliberately NOT the SMARTHOME_MQTT_BROKER setting, and not configurable. That variable
+# is the address the *device* dials to reach this machine (192.168.1.238 here); pointing
+# host-side clients at it would send them out over the LAN to reach a broker on the same
+# box, and break the moment the DHCP lease changes. The port is genuinely shared -- both
+# sides must agree on it -- the host is not.
+$SmartHomeLocalBrokerHost = '127.0.0.1'
+
 function Get-SmartHomeDevEnvPath {
     param(
         [Parameter(Mandatory = $true)]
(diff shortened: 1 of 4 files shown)


JojoEffect:
## The hardware run failed, and it was my change

Recording this because the description above predicted a saving and did not predict this.

```
  HomieClientCheck     FAIL  40s  7 conformance failure(s):
    not retained: homie/homie-client-check/$state
    not retained: homie/homie-client-check/matrix/counter/$name
    not retained: homie/homie-client-check/matrix/counter/$datatype
    not retained: homie/homie-client-check/matrix/lifecycle/$name
    not retained: homie/homie-client-check/matrix/lifecycle/$datatype
    not retained: homie/homie-client-check/matrix/counter/$retained
    not retained: homie/homie-client-check/matrix/counter/$settable
```

The bounded poll ended the wait once the subscriber's output had been quiet for 250ms, on the stated assumption that the retained replay arrives as **one contiguous burst**. It does not — the output is buffered and lands in chunks with gaps wider than that. The snapshot was cut short mid-replay, and every topic that had not arrived yet was scored as missing.

**Why my verification missed it.** The harness compared the new code's topic count against itself across four runs. That is self-consistency, not correctness — it never once compared against the old behaviour, so a systematically truncated snapshot looked perfectly stable. A harness that cannot disagree with itself proves nothing, and this is the second time in this session that a green-looking check turned out to be measuring the wrong thing.

## What changed

A fixed window again, but **1 second rather than 3**. Two of those three seconds were the IPv6 connect timeout removed in the first commit, not replay time; with the connect at ~0.02s the whole store lands well inside a second. Deterministic, and easy to reason about, which is worth more than shaving another half second by guessing when a buffered stream has ended.

The `127.0.0.1` change is untouched and was never implicated — the failure was entirely in the polling heuristic.

Expected result now: roughly 3.00s → 1.00s per snapshot rather than the 0.77s claimed above. Re-running the suite to confirm, since a prediction is exactly what got this wrong the first time.

JojoEffect:
## Final hardware run — 5/5, and the timing table in the description is now wrong

**Supersedes the "Testing Done" numbers above.** A cleanup pass over this branch's own diff found the cause of the slowdown I had accepted as unavoidable, so the cost is no longer there to report.

```
  WifiCheck            PASS   30s
  MqttCheck            PASS   28s
  Bmp280Check          PASS   30s
  HomieClientCheck     PASS   59s
  MqttReconnectCheck   PASS   54s
  TOTAL                      201s
```

| | baseline | after retain fix | now |
|---|---|---|---|
| HomieClientCheck | 62–64s | 74s | **59s** |
| suite total | 195–206s | 214s | **201s** |

So the final position is better than either previous claim: the latent test bug is fixed **and** the suite is no slower than before — `HomieClientCheck` is in fact a few seconds faster than baseline.

### What the 15 seconds were

The four lifecycle waits (`alert`, `ready`, `sleeping`, `ready`) read only `Ok`/`Seen` and discard the snapshot, but they were inheriting the new retain requirement. The device republishes `$state` within milliseconds of a `/set`, and the snapshot subscriber now connects ~150–250ms later — so whether that value arrived *replayed* or *live* was a coin flip, and every loss cost a whole extra 3s snapshot to re-observe an identical value.

`Wait-ForRetainedValue` now takes `-RequireRetained`, default on, turned off at exactly those four sites. The two callers that consume `.Snapshot` keep it, because that is what makes reusing the snapshot sound.

Nothing is given up: `$state`'s retained-ness is still asserted once by `Test-Attribute` against the announce snapshot. Re-proving it four more times added no coverage.

### Also in that pass

- **The loopback address is a named constant.** Three of four review angles independently flagged that it had become shared vocabulary with four spellings and no constant — two of them message strings a grep for `-h 'localhost'` would never find. `$SmartHomeLocalBrokerHost` now sits with `$SmartHomeDevEnvPrefix` and `$SmartHomeHomieTopic`, under the header that already states the rule.
- **The settle window is a constant, not a parameter**, since both call sites had already dropped their overrides and its own comment says the value is permanent.
- **Its justification was self-contradictory.** It argued the mid-announce interleaving "is not one this code can observe" — which the retain-flag commit in this same branch made false. Corrected to say what it actually still protects: the callers that read a snapshot *without* a flag guard.
- **The per-caller retain policy is written down** where the four callers share code. One documenting it and three not was an invitation to "fix" the `/set` loop by adding the same guard, which would make it slower and flaky against live echoes.
- `local.env.template.ps1` claimed the broker host was "used by the device and the local subscriber" — false since the first commit here. Corrected, including why it is deliberately not configurable.

Issue #20, opened 2026-08-22
Test name is spelled in three places, and WRONG-TEST exists to catch the drift


JojoEffect:
A test's name is spelled in three independent places: the folder and `.nfproj` name, the
key in `$testCatalog`, and the device's own `TestName` const. Nothing in the build
reconciles them, so the runner grew a `WRONG-TEST` outcome to report their disagreement
— a naming-convention slip surfaced as a device-level verdict, on hardware, 90 seconds
into a run, in the same shape as the stale-deploy failure it has to be distinguished
from.

`Common.ps1`'s `Get-NfProjectAssemblyName` already reads `<AssemblyName>` (which is
`SmartHome.IntegrationTests.<Name>`) for exactly this class of problem in
`Deploy-ToDevice.ps1`. Deriving the expected marker prefix from the project instead of
the catalog key would let both the per-project `TestName` const and the `WRONG-TEST`
outcome disappear.

The marker *grammar* being split across PowerShell and C# is unavoidable and fine; it is
the name that should not be.

**Files:** `scripts/Run-IntegrationTests.ps1`,
`src/integrationTests/*/Program.cs`

Issue #12, opened 2026-08-22
OvenControl is an empty stub


JojoEffect:
`src/devices/OvenControl` builds and does nothing.

**Blocked on the same question as IrrigationControl:** what is physically wired — a
relay, GPIO directly, a thermocouple over SPI/I2C, or nothing yet?

Follow RoomSensor's shape once known: `NetworkHelper.ConnectToConfiguredNetwork()`,
build with `HomieDeviceBuilder`, `IHomieClient.ConnectWithRetry()`, and act on
`IHomieClient.OnCommand` rather than `property.OnUpdate`.

An oven is the case where the `OnCommand` distinction matters most: a heater started
twice because the device re-consumed its own retained publish is a safety problem, not
a cosmetic one.

**Files:** `src/devices/OvenControl/`

Issue #11, opened 2026-08-22
IrrigationControl is an empty stub


JojoEffect:
`src/devices/IrrigationControl` builds and does nothing.

**Blocked on one answer:** what is physically wired to the board for irrigation — a
relay module, GPIO pins directly, something over I2C, or nothing yet? Everything else
follows from that.

Once known, follow RoomSensor's shape:

- `NetworkHelper.ConnectToConfiguredNetwork()` — the only WiFi path in this repo.
- Build the device with `HomieDeviceBuilder`, then `IHomieClient.ConnectWithRetry()`.
  The Homie client must own the MQTT session; connecting the transport first produces a
  session with no last will.
- Act on `IHomieClient.OnCommand`, **not** `property.OnUpdate`. The property event fires
  both when a controller sets a value and when the device updates its own, and cannot
  tell them apart. For an actuator that distinction is the difference between opening a
  valve once and opening it twice.

**Files:** `src/devices/IrrigationControl/`

PR #32, opened 2026-08-23
fix(homie): reflect the lifecycle outcome, and make the conformance check measurable

### LINKED ISSUE #13: Homie lifecycle vocabulary is duplicated instead of derived from the State enum
`HomieClientCheck` spells `ready` / `alert` / `sleeping` in its own consts, again in its
hand-built `$format` string, and `Run-IntegrationTests.ps1` spells them a third time —
while `Enums/State.cs` and `StateExtensions.GetString()` already own that vocabulary.
Adding a state means editing three places.

Related, and the more interesting half: `HomieClient.HandleIncomingMessage` calls
`property.Set(message)` before raising `OnCommand`, so the property reflects the command
optimistically. A refused transition (`Alert -> Sleeping` is illegal per
`Device.CanChangeState`) therefore leaves the retained store showing
`matrix/lifecycle=sleeping` next to `$state=alert`. The conformance check only asserts
`$state`, so it passes.

A purpose-built reference device that publishes a state it is not in is a poor pattern
for the actuators that will copy it.

Suggested: build the `$format` string from the `State` enum, and update the lifecycle
property from the *result* of `Alert()` / `Sleep()` / `Ready()` rather than letting
`Set()` reflect the command.

**Files:** `src/integrationTests/HomieClientCheck/Program.cs`,
`scripts/Run-IntegrationTests.ps1`

### LINKED ISSUE #34: Refused-transition step cannot tell a refusal from a lost command
The Homie conformance check drives one deliberately illegal transition — `alert -> sleeping`, which `Device.CanChangeState` forbids — and asserts that neither `$state` nor the `lifecycle` property moved off `alert`. That assertion is what guards the "reflect the outcome, not the command" fix in #32.

It cannot currently tell a refusal from a command that never arrived.

## Why

The step publishes `sleeping` to `matrix/lifecycle/set` once, takes one settled snapshot, and checks two topics:

```powershell
$refused = Get-HomieRetainedSnapshot -Port $Port
# asserts $state == 'alert' and matrix/lifecycle == 'alert'
```

A `/set` is non-retained, as the convention requires of a controller. If it is dropped — the device not yet subscribed, a reconnect in the gap — nothing reaches the device, nothing changes, and both assertions pass. A lost command and a correctly refused command are indistinguishable at the snapshot.

Unlike the four steps around it, this one cannot simply retry its way out: the other steps poll until `$state` reaches a new value and now republish each round (`Wait-ForRetainedValue -RepublishTopic`), whereas this step's whole assertion is that nothing changes, so there is no echo to wait on.

This is not hypothetical in the sense that matters: run 3 of the eight hardware runs on 2026-08-26 lost all five `/set` commands of the preceding phase, which is the same mechanism landing on a different step.

## What keeps it honest today

The step immediately before it drives the device to `alert` and waits for `$state` to confirm — so the device demonstrably applied a `/set` on this very topic seconds earlier. The window for it to go deaf in between is small. Small is not zero, and the comment at the call site says so rather than leaving it for the next reader to find.

## Suggested fix

Assert arrival from the live `homie/#` log instead of inferring it from the snapshot.

`HomieClient` reflects the `/set` payload onto the property before the app sees it, and `HomieClientCheck.HandleCommand` then publishes the real state over it. So a refused `sleeping` command produces two messages on `matrix/lifecycle`, in order:

1. `sleeping` — the library's optimistic reflection
2. `alert` — the device's correction

Both are in the live log. Asserting that the pair appears *after* this step's publish proves the command arrived, proves the reflection happened, and proves the correction overwrote it — which is a strictly stronger measurement than "the store settled on `alert`", and is the actual behaviour #32 fixed.

Mechanically: record the live log's line count before publishing, then look only at lines added after it. The log is cumulative from broker start and `sleeping` also appears in the legitimate sleeping step later, so an unanchored search would match the wrong message.

`$extensions` is already asserted from the live log rather than the retained store (for an unrelated reason — an empty retained payload is a delete in MQTT), so reading that log mid-check is established practice here, not a new mechanism.

## Files

- `scripts/Run-IntegrationTests.ps1` — the `Refused` branch of `$lifecycleSteps` in `Invoke-HomieConformanceCheck`
- `src/integrationTests/HomieClientCheck/Program.cs` — `HandleCommand`, which produces the correction being asserted

## Verification

Needs a hardware run (`scripts\Run-IntegrationTests.ps1 -Tests HomieClientCheck`). Worth confirming the assertion actually fails when it should: temporarily reverting `HandleCommand`'s `_lifecycle.Update(...)` should make the new check fail and the old one pass, which is the difference the issue is about.

JojoEffect:
Closes #13.

Also closes #34.

Started as the `lifecycle` reflection fix. Running it on hardware turned up three further
defects in the conformance check itself, and the timing question the first hardware run
raised is now answered with measurements rather than arithmetic. Six commits, each standing
on its own.

## 1. Reflect the outcome, not the command

`HomieClient.HandleIncomingMessage` calls `property.Set(message)` before raising
`OnCommand`, so a settable property reflects the command optimistically. That is correct
for an ordinary property — a controller writes `21.5`, the device publishes `21.50` back —
but the `lifecycle` property's value is a *request*, and a request can be turned down:

- `Device.CanChangeState` forbids `alert -> sleeping` (alert may only return to `ready`, or disconnect)
- an unknown payload is not a state at all

In both cases the retained store was left advertising `matrix/lifecycle=sleeping` next to
`$state=alert`. Both are retained, so this is not a transient that goes by on the wire —
every controller that connects afterwards reads the contradiction out of the broker's
store. The conformance check only asserted `$state`, so it passed.

`HandleCommand` now publishes the device's actual state over the reflection once the
command has been applied or refused. `CLAUDE.md` carries the rule as the third thing to
know when writing an actuator:

> Reflect the outcome, not the command. […] Publish the real value over it once the
> command has been applied or refused, or the retained store advertises a state the device
> is not in, to every controller that connects afterwards.

`HomieClientCheck` is the reference device the actuators (#11, #12) are meant to copy,
which is what makes this worth fixing before they exist rather than after.

The lifecycle vocabulary now comes from the `State` enum. `StateExtensions.GetString()`
already owns those names; `HomieClientCheck` had its own three consts, spelled them again
into the `$format` string, and the runner spells them a third time. One array is now the
single source for both the `$format` string and the set of commands the dispatch accepts,
and the host asserts `$format` still reads `ready,alert,sleeping` — so if the derivation
breaks, that is what notices. The PowerShell/C# boundary keeps its own copy, deliberately:
the runner cannot call into the device. That is the same unavoidable split #20 notes for
the `[ITEST]` marker grammar.

The suite now asks for `alert -> sleeping` **on purpose** — the one transition the drive
sequence previously went out of its way to avoid — and asserts that neither `$state` nor
the lifecycle property moved. The five drive steps became a table so the refusal sits
between `alert` and `ready` instead of needing its own `alert` / `ready` round trip.

## 2. A retransmitted duplicate must not erase a proved retain flag

The first hardware run failed, and not because of the change above. Three attributes of 53
came back as "not retained": `color-value/$name`, `color-value/$datatype`,
`lifecycle/$format`. A steady-state probe then showed all three retained, six snapshots out
of six, so the store was never the problem.

A 40s capture of one re-announce shows what is: a retransmitted tail of the last property's
attributes, and counter values arriving out of order — 146 after 147. Nothing but
retransmission produces that. Confirmed in the sibling checkout rather than assumed:
`MqttClient.cs:1919` sets `DupFlag = true` when it resends an in-flight message, up to
`MqttSettings.MaximumAttemptsRetry` (3).

`Get-HomieRetainedSnapshot` kept the last message per topic, so a live duplicate arriving
*after* the replayed copy overwrote it with `retain=0`. The attribute was in the store the
whole time; the evidence that it was got thrown away.

The merge rule is now: a repeat of the **same** payload only adds to what is known about a
topic. A **different** payload still replaces both fields — otherwise a value delivered
only live would inherit the retained-ness of the value it replaced, which is precisely the
bug #31's flag check exists to catch. That is why this is not simply an `-or` over the flag.

## 3. Phase timing, and what the "unexplained 14s" actually was

The first hardware run took 73s against a 59s baseline, and the suite reported one number
per test, so there was nothing to attribute the difference to. The check is now timed by
phase, with snapshots counted per phase and the deploy separated from the measurement:

```
  phase breakdown (3s per snapshot):
    announce                 10.1s   3 snapshot(s)
    attributes                0.0s   0 snapshot(s)
    /set round-trip           3.4s   1 snapshot(s)
    lifecycle                16.8s   5 snapshot(s)
    re-announce              12.9s   2 snapshot(s)
```

Almost all of this check's wall clock is 3s snapshot windows, and the poll loops take an
unpredictable number of them, so snapshots-per-phase is the number worth having next to the
seconds.

Eight runs on 2026-08-26 settle the question. A healthy check is **43.1–44.1s, a spread
under one second across seven runs**. Deploy is 16–29s and swings independently. So the 73s
was an ordinary 43s check behind a slow flash — the delta was never in the check, and the
single figure could not show that.

The `+3s` estimate for the refusal step was right, and is now checkable rather than
asserted: lifecycle costs 5 snapshots where it cost 4, and a snapshot is 3.4s.

Phase seconds render in the invariant culture. The first run printed `16,7s` on this
machine, which reads as a thousands separator to everyone who was not sitting at it — and
these lines exist to be quoted into pull requests and issues.

## 4. Retry a controller command instead of timing out on a lost one

One of those eight runs failed, and it exposed a defect in the check that predates this
branch. The check published each `/set` exactly once and then only polled for the echo. A
`/set` is non-retained, as the convention requires of a controller, so a command reaching
the broker while the device is not yet subscribed is dropped outright — nothing remains in
the store for the device to collect when it does subscribe. That single lost message became
the full `CommandTimeoutSeconds` of polling for an echo that could never arrive, and five
conformance failures against a device that was working.

The failing run was the only one of the eight whose announce wait was satisfied by its
first snapshot rather than its third — the only one, in other words, that reached the `/set`
phase early. Its `/set` phase spent 33.4s over 10 snapshots with all five properties still
holding their boot values, while the lifecycle commands thirty seconds later were applied
normally.

Commands are now published at the top of every poll round rather than once before the loop,
and only for properties still pending. `Wait-ForRetainedValue` gains the same option, so the
four lifecycle steps that wait for a state transition retry their command too; the announce
and re-announce waits pass nothing and are unaffected.

Retrying is the only fix available at this layer. MQTT gives a publisher no signal about who
is subscribed, and `$state=ready` is the device's claim about itself rather than about the
broker's routing table, so the host cannot wait for the condition it actually needs. On a
healthy run this costs nothing: the first round succeeds and no command is ever re-sent.

## 5. Prove the refused transition from the wire, not from where it settled

Closes #34, which the retry in section 4 exposed rather than caused. The refused step read
a settled snapshot, and a command that never arrived leaves exactly the snapshot a refused
one does — so a lost `/set` passed it as a refusal. Unlike the polls around it, this step
cannot retry its way out: its whole assertion is that nothing changes, so there is no echo
to wait on.

The difference is on the wire. A refusal that actually happened puts `sleeping` (the
library's optimistic reflection) and then `alert` (the device's correction) on the property
topic, in that order. A command that never arrived puts neither. The step now opens its
capture window before publishing and reads that sequence, with the retained replay excluded
— the replayed copy predates the command and says nothing about whether it arrived.

Three failures now separate what used to be one silent pass: the command never reaching the
property, the reflection being left standing, and the correction arriving before the
reflection rather than over it. The middle one is the defect section 1 fixes.

`Get-HomieRetainedSnapshot` is split into `Start-HomieCapture`, `Stop-HomieCapture` and
`ConvertTo-HomieSnapshot`, composing to what it was for its existing callers. That split is
what lets this step publish *inside* the window. `Start-HomieCapture` also gains
`-WaitForConnectSeconds`, because `Start-Process` returns before `mosquitto_sub` has a
session and a publish issued immediately after could otherwise beat the subscriber onto the
broker — which would read as "the command never arrived" and fail a healthy device.

## Known limits, filed rather than fixed here

- Why the device missed those five commands is not established. The library subscribes
  before it announces, and both travel the same ordered connection, so at the broker the
  SUBSCRIBE precedes the `$state` publish — which argues against the simple "published too
  early" story. The retry removes the symptom without proving the cause. Filed, along with
  the reason it was hard to chase: conformance runs capture no device log.

## Considered and not done

Making the library itself defer the reflection until the app confirms the command. That
would mean `OnCommand` returning an accept/reject result, which is a real API change to
`IHomieClient` and a design decision rather than a bug fix. The device-level correction is
also the more honest layering: only the app knows whether a command succeeded.

## Testing done

- CI green on every commit.
- **16 `HomieClientCheck` runs on real hardware** (ESP32 on COM3, local Mosquitto):
  - 8 before the retry fix — 7 PASS, 1 FAIL (the FAIL is analysed in section 4)
  - 3 after the retry fix — 3 PASS, check at 43.4s, 50.0s and 43.5s
  - 2 after the wire assertion — PASS, check at 44.1s and 43.3s, snapshot counts unchanged
  - 2 negative runs, each reverted afterwards, one per failure mode section 5 adds:

| injected defect | result |
|---|---|
| `_lifecycle.Update()` commented out in `HandleCommand` — the #13 defect itself | FAIL, **2** failures: the settled assertion, plus `device left the reflected 'sleeping' on matrix/lifecycle and never published 'alert' over it` |
| the refused step's `/set` published to a topic nothing subscribes to — a lost command | FAIL, **1** failure: `refused 'sleeping' never reached matrix/lifecycle -- the command was lost, so nothing about the refusal was measured` |

  The second run is the one that matters for #34. Both settled assertions *passed* — `$state`
  was `alert`, the property was `alert` — so before this change that run would have been a
  clean PASS with the command never arriving. That is the vacuous pass #34 describes,
  reproduced on hardware and now caught, with nothing else in the check reacting to it.
- The three passes after the retry fix cannot prove a one-in-eight race fixed, and are not
  offered as that. They show no regression; the retry stands on the mechanism rather than
  on the sample.
- The positive runs also settle a doubt about section 5's mechanism: the assertion depends
  on the tail of a capture reaching disk, where every existing use only needed the bulk
  retained replay, and a buffered tail lost to the process-tree kill would have failed a
  healthy device. It does not.
- Section 5's third failure mode — the correction arriving *before* the reflection rather
  than over it — is not exercised on hardware. Producing it means reordering two publishes
  inside `HandleCommand` against the library's own ordering, which is a contrived device
  rather than a defect anyone would write. The ordering comparison itself was checked in
  isolation.
- The other four integration tests were not re-run. Only `HomieClientCheck` reaches
  `Invoke-HomieConformanceCheck`; the rest are `DeviceMarker` or `BrokerOutage` and read the
  device log or the live `homie/#` log instead.
- RoomSensor has been redeployed, so the device is back to doing its real job.

### DIFF
--- a/.claude/skills/smarthome-integration-tests/SKILL.md
+++ b/.claude/skills/smarthome-integration-tests/SKILL.md
@@ -16,8 +16,9 @@ Run `scripts\Run-IntegrationTests.ps1` — one call covers every project under
 - **`HomieConformance`** (HomieClientCheck) — the host measures a purpose-built device against
   the Homie v4 convention: mandatory attributes and their retained flags, one property of every
   datatype with its `$format`/`$unit`/`$settable`/`$retained`, a `/set` command applied and
-  reflected back, the `alert` and `sleeping` states driven through a control property, and a full
-  re-announce after the broker is replaced. It emits no `[ITEST]` marker by design — the evidence
+  reflected back, the `alert` and `sleeping` states driven through a control property, a refused
+  transition that must not be advertised as if it had happened, and a full re-announce after the
+  broker is replaced. It emits no `[ITEST]` marker by design — the evidence
   is what lands on the broker.
 
 Both host-decided kinds (`BrokerOutage`, `HomieConformance`) take the broker away and put a fresh
--- a/.claude/skills/smarthome-integration-tests/SKILL.md
+++ b/.claude/skills/smarthome-integration-tests/SKILL.md
@@ -49,6 +49,14 @@ Reading the result:
 - **exit 1** — the summary names each non-passing test and prints its captured device log path.
   Start there.
 
+Reading the timing: each summary line is `<total>s (<deploy>s deploy)`. The deploy is a full
+`/t:Rebuild` plus a flash and swings by tens of seconds independently of the test, so a run that
+looks slower is usually a build that was — compare the *difference*, not the total.
+`HomieConformance` additionally prints a per-phase breakdown (announce, attributes, `/set`
+round-trip, lifecycle, re-announce) with the number of 3s retained-store snapshots each phase
+took. That count is decided at runtime by how many polls a step needs, so it is the figure that
+explains a conformance run's duration; nothing about it is derivable from reading the script.
+
 Outcomes other than PASS/FAIL:
 
 - `NO-RESULT` — no marker within the capture window. The app crashed before reporting, or wasn't
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -182,7 +182,7 @@ device owns a connection rather than being one, and exposing `Publish`/`Subscrib
 let an app publish an attribute non-retained or `$state` out of order. The `Device` model (built
 with `HomieDeviceBuilder`) says what the device *is*; `IHomieClient` is what you *do* with it.
 
-Two things to know when writing an actuator (Irrigation, Oven):
+Three things to know when writing an actuator (Irrigation, Oven):
 
 - Act on `IHomieClient.OnCommand`, not on `property.OnUpdate`. The property event fires both
   when a controller sets a value and when the device updates its own, and cannot tell them
@@ -190,6 +190,13 @@ Two things to know when writing an actuator (Irrigation, Oven):
 - Settable properties are subscribed on `homie/[device]/[node]/[property]/set`, as the spec
   requires. Until 2026-08-21 the code subscribed to the property *value* topic instead, so
   commands were never received and the device re-consumed its own retained publishes.
+- Reflect the outcome, not the command. `HomieClient` publishes a `/set` payload onto the
+  property before `OnCommand` runs, which is right for an ordinary property and wrong for one
+  whose value is a *request* the device can turn down — an illegal `$state` transition, an out
+  of range setpoint, a relay that failed to close. Publish the real value over it once the
+  command has been applied or refused, or the retained store advertises a state the device is
+  not in, to every controller that connects afterwards. `HomieClientCheck` does this for its
+  `lifecycle` property and the conformance suite asserts it.
 
 `HomieClient` owns its MQTT session and must: Homie v4 requires the connection to carry a last
 will setting `homie/[device-id]/$state` to `lost`, and a will can only be declared in CONNECT.
@@ -235,8 +242,9 @@ The suite runs three kinds of test, and the kind is declared per entry in
 - **`HomieConformance`** — the host measures a purpose-built device against the Homie v4
   convention: mandatory attributes and their retained flags, one property of every datatype with
   its `$format`/`$unit`/`$settable`/`$retained`, a `/set` command applied and reflected back, the
-  `alert` and `sleeping` states driven through a control property, and a full re-announce after
-  the broker is replaced. `HomieClientCheck` is this kind. Retained-ness is read from a *fresh*
+  `alert` and `sleeping` states driven through a control property, a refused transition that
+  must not be advertised as if it had happened, and a full re-announce after the broker is
+  replaced. `HomieClientCheck` is this kind. Retained-ness is read from a *fresh*
   subscriber (`mosquitto_sub -F '%t %r %p'`), because MQTT only sets the retain flag when
   replaying from the store — a live retained publish arrives with the flag clear.
 - **`BrokerOutage`** — the host decides. The device publishes a heartbeat and subscribes to an
--- a/scripts/Run-IntegrationTests.ps1
+++ b/scripts/Run-IntegrationTests.ps1
@@ -501,6 +501,11 @@ function Get-MosquittoTool {
 # of *listening*, where they used to be two of connecting and one of listening.
 $SnapshotSettleSeconds = 3
 
+# Snapshots taken so far, across the whole run. Declared here rather than left to the
+# first ++ because Set-StrictMode -Version Latest makes incrementing an unset variable a
+# terminating error, which would arrive as an 'ERROR' verdict blamed on the device.
+$script:snapshotsTaken = 0
+
 function Get-HomieRetainedSnapshot {
     # What a controller joining *now* would see: the broker's retained store.
     #
@@ -539,6 +544,12 @@ function Get-HomieRetainedSnapshot {
 
     Start-Sleep -Seconds $SnapshotSettleSeconds
     Stop-SmartHomeProcessTree -ProcessId $process.Id
+
+    # Counted, not estimated. Nearly the whole wall clock of a conformance run is spent
+    # in this function, and the poll loops that call it decide at runtime how many times
+    # they need to -- so the number of snapshots a run actually took is the one figure
+    # that explains its duration, and it cannot be read off the code.
+    $script:snapshotsTaken++
 
     # Last message per topic wins -- except that a repeat of the SAME payload only adds
     # to what is known about it.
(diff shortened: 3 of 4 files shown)


JojoEffect:
## Hardware run: PASS, 73s

```
  HomieClientCheck     PASS   73s   attributes, datatypes, retained flags, /set round-trip,
                                    alert/sleeping, a refused transition and re-announce all conform
```

`retained store holds 53 topics`, which matches the steady-state probe exactly. The failing run before the second commit reported 54 with three of them carrying `retain=0` — that gap is gone.

### The first run failed, and it was not this change

Three attributes of 53 came back as "not retained": `color-value/$name`, `color-value/$datatype`, `lifecycle/$format`. A steady-state probe then showed all three retained, six snapshots out of six, so the store was never the problem.

A 40s capture of one re-announce shows what is: a retransmitted tail of the last property's attributes, and counter values arriving out of order — 146 after 147. Nothing but retransmission produces that. Confirmed in the sibling checkout rather than assumed: `MqttClient.cs:1919` sets `DupFlag = true` when it resends an in-flight message, up to `MqttSettings.MaximumAttemptsRetry` (3).

`Get-HomieRetainedSnapshot` kept the last message per topic, so a live duplicate arriving *after* the replayed copy overwrote it with `retain=0`. The attribute was in the store the whole time; the evidence that it was got thrown away.

The merge rule is now: a repeat of the **same** payload only adds to what is known about a topic. A **different** payload still replaces both fields — otherwise a value delivered only live would inherit the retained-ness of the value it replaced, which is precisely the bug #31's flag check exists to catch. That is why this is not simply an `-or` over the flag.

Two of the three topics are ones the suite has always asserted, so this was reachable before this branch. Asserting `lifecycle/$format` widened the sample by one.

### Timing: worse than I estimated

73s against a 59s baseline. I predicted ~62s on the grounds that the refusal step adds one snapshot (~3s), and that does not account for 14s. One sample, and the run before it took 63s while failing early — so I would not read a trend into it either way. Every `Wait-ForRetainedValue` poll is a 3s snapshot, and the step table has five steps where the old loop had four, so a single extra poll landing badly costs more than the arithmetic suggests. Say the word if you want it measured properly rather than left as an unexplained delta.

### Scope

Only `HomieClientCheck` touches `Get-HomieRetainedSnapshot` / `Wait-ForRetainedValue` — every caller sits inside `Invoke-HomieConformanceCheck`. The other four tests are `DeviceMarker` or `BrokerOutage` and read the device log or the live `homie/#` log instead, so they are unaffected by the second commit and were not re-run.

The device is left holding `HomieClientCheck`; RoomSensor needs a redeploy before it does its real job again.

Issue #35, opened 2026-08-28
HomieClientCheck lost all five /set commands in 1 run of 8, and left no evidence


JojoEffect:
One `HomieClientCheck` run in eight on 2026-08-26 lost all five `/set` commands of the round-trip phase. #32 makes the check retry them, which removes the symptom. It does not establish why the device missed them, and the retry would equally mask a real device-side defect.

## What was observed

Eight consecutive runs, same image, same broker, same machine. Seven passed identically. The one failure differed in exactly one respect before it failed:

| | 7 passing runs | the failing run |
|---|---|---|
| announce phase | 9.9–10.3s, **3 snapshots** | 3.4s, **1 snapshot** |
| retained store | 53 topics | 52 topics |
| `/set` phase | 3.4–3.5s, 1 snapshot | **33.4s, 10 snapshots** (full timeout) |
| `/set` outcome | all five echoed | all five still at boot values |
| lifecycle phase | 16.7–17.2s, 5 snapshots | 16.7s, 5 snapshots — **passed normally** |

The properties still held `0`, `0.00`, `false`, `low` and `initial`. Thirty seconds later the lifecycle `/set` commands on the same node were applied without trouble, so the device was alive, connected and processing commands by then.

A 1-snapshot announce means the host's first snapshot already saw `$state=ready` **retained**, i.e. replayed from the store — so the device had finished announcing before that subscriber even connected. The host therefore reached the `/set` phase roughly 7s earlier than in the passing runs.

## Why "published too early" does not obviously explain it

`HomieClient.Connect()` subscribes to the settable `/set` topics *before* it announces, deliberately, with a comment saying why:

```csharp
SubscribeSettablePropertyTopics();

// Subscriptions before the announcement, so a controller reacting to it
// cannot find the device deaf to /set.
if (!Announce(State.Ready))
```

SUBSCRIBE and the subsequent `$state` publish travel the same ordered TCP connection, so at the broker the subscription exists before `$state=ready` is stored. A controller that reacts to a *retained* `$state=ready` should therefore never find the device unsubscribed.

Something breaks that reasoning. Candidates, none verified:

- the device reconnected between announcing and the commands arriving, and the commands fell into the gap (`ReconnectingMqttClient` would re-announce via `HandleConnectionOpen`, which would also restore the store, so the host would see nothing amiss)
- a QoS/retransmission interaction — the retained-flag defect fixed in #32 was in the same area
[...]

Issue #21, opened 2026-08-22
Test kind is dispatched by if/elseif rather than by capability


JojoEffect:
`Run-IntegrationTests.ps1` branches on `$settings.Kind` in several places: the
`-NoBroker` guard is a two-term OR over kinds, and the verdict dispatch is a pair of
branches that differ only in which function they call.

What the host-decided kinds actually share is "owns the broker's lifetime". What differs
is only which function returns the verdict. Adding a fourth kind therefore means several
coordinated edits, and the `-NoBroker` list is the one that gets forgotten — silently,
because the result is a confusing failure rather than an error.

Catalog entries could carry `OwnsBroker = $true` and `Verdict = '<function name>'`
instead of a `Kind`, leaving one dynamic invoke and one shared epilogue. Not speculative:
three kinds exist today and two of them already differ only in the function name.

**Files:** `scripts/Run-IntegrationTests.ps1`

PR #40, opened 2026-08-29
feat(scripts): keep the evidence a conformance failure has to be read from

JojoEffect:
Closes the evidence half of #35. The intermittent lost-command failure itself is **not** touched here and #35 stays open for it.

## The gap

A `HomieConformance` run left nothing behind to investigate:

- Managed debug output was captured only for `DeviceMarker` tests. `$logFile` was written for every other kind and stayed empty for this one, so the conformance path never produced a device log at all.
- Both broker-side logs are transient by design. `Start-DevEnv.ps1` truncates them on every start, `Stop-DevEnv.ps1` deletes them on every stop, and the host-decided checks cycle the broker mid-test — so the wire evidence for a verdict is gone before anyone reads the verdict.

When `HomieClientCheck` lost all five of its `/set` commands in one run of eight, the summary line was all there was.

## The change

All of it in `scripts/Run-IntegrationTests.ps1`.

- **Managed debug output across a conformance run.** `Start-DeviceDebugCapture` / `Stop-DeviceDebugCapture` run `Watch-DeviceDebugOutput.ps1` alongside the check and write `<test>.log`, the same file the `DeviceMarker` path already produced.
- **Broker and `homie/#` logs preserved per generation.** `Save-BrokerEvidence` copies both into the run's `$LogDirectory` before every broker stop, named `<test>-NN-broker.log` / `<test>-NN-homie.log`. It runs from `Stop-SuiteBroker`, which is the single funnel every stop and restart goes through, plus once more in the suite's own teardown.

Three decisions worth stating:

- The capture runs with **`-NoReboot`**, unlike the `DeviceMarker` captures. Those reboot because a missed boot is indistinguishable there from a device that never reported, and their verdict is read out of the log. Here the verdict comes from the broker, so missed lines cost only evidence — while a reboot would restart the announce the check is about to measure, which is the timing #35 is about.
- Its duration is a **backstop, not a wait**, derived from the check's own timeouts (`Get-ConformanceCaptureSeconds`, 420s at the current catalog values). The capture is closed as soon as the check returns.
- **Neither addition can change a verdict.** A monitor that cannot attach warns and leaves an empty log. A log that cannot be copied warns and is skipped. Stopping the capture kills the process tree only when it is still running — `taskkill` on a dead pid writes to stderr, and under this script's `$ErrorActionPreference = 'Stop'` that is a terminating `NativeCommandError` thrown out of the `finally` that stops the capture, which would replace the verdict the check had already reached. That one is measured, not assumed.

#20 and #21 propose restructuring the same verdict-dispatch area. The dispatch itself grows by eight lines; everything else is new functions above it.

## Verified on hardware

Full suite on the real ESP32, all five green, and the summary now names what else is in the log directory:

```
  WifiCheck            PASS         22s (17s deploy) connected to the configured WiFi network
  MqttCheck            PASS         20s (15s deploy) round-trip through 192.168.1.238 on smarthome-test/echo
  Bmp280Check          PASS         19s (14s deploy) 24.262009808258154°C, 942.05946376109111hPa, 54.315140903821391%RH
  HomieClientCheck     PASS         64s (18s deploy) attributes, datatypes, retained flags, /set round-trip, alert/sleeping, a refused transition and re-announce all conform
  MqttReconnectCheck   PASS         50s (15s deploy) republished and stayed subscribed across outages of 3s, 20s
  TOTAL                             175s (79s deploy)
```

What the run left behind — `HomieClientCheck.log` is the file that did not exist before, and `MqttReconnectCheck` contributes a generation per outage:

```
   10474  Bmp280Check.log
  127497  HomieClientCheck-01-broker.log
    5927  HomieClientCheck-01-homie.log
   20634  HomieClientCheck.log
    8657  MqttCheck.log
   47188  MqttReconnectCheck-02-broker.log
    3992  MqttReconnectCheck-02-homie.log
    1537  MqttReconnectCheck-03-broker.log
      51  MqttReconnectCheck-03-homie.log
    1537  MqttReconnectCheck-04-broker.log
      51  MqttReconnectCheck-04-homie.log
    5650  MqttReconnectCheck-05-broker.log
     706  MqttReconnectCheck-05-homie.log
    1510  WifiCheck.log
```

Checked against the three questions #35 says a repeat of the failure would need answered:

**Did the device log a reconnect** — `HomieClientCheck.log`:

```
MQTT connection lost. Reconnect handler called.
Start reconnect...
Trying to reconnect once...
Reconnect attempt finished with result: 0.
```

**Did the command handler ever fire** — one line per `/set`, same file:

```
HomieClientCheck: command on 'string-value' -> 'commanded'
HomieClientCheck: command on 'enum-value' -> 'high'
HomieClientCheck: command on 'boolean-value' -> 'true'
HomieClientCheck: command on 'float-value' -> '21.5'
HomieClientCheck: command on 'integer-value' -> '42'
```

**What the broker did with the five PUBLISH packets** — `HomieClientCheck-01-broker.log`, which holds 702 PUBLISH lines including:

```
Received PUBLISH from auto-22A052D9-... (d0, q0, r0, m0, 'homie/homie-client-check/matrix/string-value/set', ... (9 bytes))
Received PUBLISH from auto-3D2E5C5E-... (d0, q0, r0, m0, 'homie/homie-client-check/matrix/enum-value/set', ... (4 bytes))
Received PUBLISH from auto-CBC8FE36-... (d0, q0, r0, m0, 'homie/homie-client-check/matrix/boolean-value/set', ... (4 bytes))
Received PUBLISH from auto-12815759-... (d0, q0, r0, m0, 'homie/homie-client-check/matrix/float-value/set', ... (4 bytes))
Received PUBLISH from auto-0CA07ECD-... (d0, q0, r0, m0, 'homie/homie-client-check/matrix/integer-value/set', ... (2 bytes))
```

The run was healthy, so this is the shape of the evidence rather than a reproduction of the failure — but it is the evidence that was missing.

Non-hardware checks alongside it: both scripts parse clean; `Save-BrokerEvidence` exercised against a live Mosquitto 2.0.22 broker log and `homie/#` subscriber log (both held open for writing by their producers, both copied); the capture start/stop exercised against a stub standing in for the monitor, confirming the arguments reach it, output streams to the log, and the tree is killed; and the dead-launcher path confirmed to warn rather than throw under `$ErrorActionPreference = 'Stop'`.

RoomSensor was redeployed afterwards, so the device is back on its real job rather than carrying `MqttReconnectCheck`.

Refs #35

### DIFF
--- a/scripts/Run-IntegrationTests.ps1
+++ b/scripts/Run-IntegrationTests.ps1
@@ -274,35 +274,48 @@ function Start-DeviceDebugCapture {
         $powerShellExe, $watchScript, $TimeoutSeconds, $LogPath
     $process = Start-Process -FilePath 'cmd.exe' -ArgumentList $command -PassThru -WindowStyle Hidden
 
-    return @{ ProcessId = $process.Id; Path = $LogPath }
+    # A record, not a bare pid. This launcher can legitimately be dead by the time the
+    # check returns -- a monitor that cannot reach the device gives up after 15s and
+    # takes its cmd.exe with it -- and minutes then pass before anything stops it, which
+    # is long enough for Windows to hand the pid to something else. That is the case
+    # New-SmartHomeProcessRecord carries a name and a start time for. -Tree because the
+    # real work is in the grandchild.
+    return @{
+        Record = New-SmartHomeProcessRecord -Label 'device debug monitor' -Process $process -Tree
+        Path   = $LogPath
+    }
 }
 
 function Stop-DeviceDebugCapture {
-    # Killed as a tree rather than asked to stop: the monitor listens for a fixed
-    # duration and has no way to be told the check is over. The COM port is released
-    # with the process, so the next test's deploy is not affected.
+    # Killed rather than asked to stop: the monitor listens for a fixed duration and has
+    # no way to be told the check is over. The COM port is released with the process, so
+    # the next test's deploy is not affected.
     param([hashtable]$Capture)
 
     if ($null -eq $Capture) {
         return
     }
 
-    # Only if it is still running. A monitor that could not attach exits on its own and
-    # takes its cmd.exe launcher with it, and taskkill then writes "process not found"
-    # to stderr -- which under this script's $ErrorActionPreference = 'Stop' is a
-    # terminating NativeCommandError, thrown out of the finally that calls this and
+    # Through the recorded-process helper, which stops the tree only when pid, name and
+    # start time all still match, and warns instead of killing when the pid has been
+    # recycled. It also covers the already-exited case: taskkill on a dead pid writes to
+    # stderr, and under this script's $ErrorActionPreference = 'Stop' that is a
+    # terminating NativeCommandError -- thrown out of the finally that calls this, and
     # replacing the verdict the check had already reached. Measured, not assumed:
     # Stop-SmartHomeProcessTree on a dead pid throws here.
-    if (Get-Process -Id $Capture.ProcessId -ErrorAction SilentlyContinue) {
-        Stop-SmartHomeProcessTree -ProcessId $Capture.ProcessId
-    }
+    Stop-SmartHomeRecordedProcess -Record $Capture.Record | Out-Null
 
-    # An attach failure must not change a verdict the broker already decided, so this
-    # warns rather than throwing -- but it says so, because an empty log is otherwise
-    # indistinguishable from a device that said nothing.
+    # A capture that recorded nothing must not change a verdict the broker already
+    # decided, so this warns rather than throwing -- but it says so, because an empty log
+    # is otherwise indistinguishable from a device that said nothing.
+    #
+    # It does not name an attach failure as the cause, because that is the one thing it
+    # cannot be: the child's stderr is redirected into this same file, so a monitor that
+    # found no device leaves its "No nanoFramework device found on ..." in here. Nothing
+    # at all on either stream means the child never got that far.
     $captured = Get-Item -LiteralPath $Capture.Path -ErrorAction SilentlyContinue
     if ($null -eq $captured -or $captured.Length -eq 0) {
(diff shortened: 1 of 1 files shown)


JojoEffect:
Re-ran the full suite on hardware against `2d49328`, after the review fixes changed how the debug capture's launcher is identified and stopped. All five green again, 174s:

```
  WifiCheck            PASS         21s (16s deploy) connected to the configured WiFi network
  MqttCheck            PASS         21s (15s deploy) round-trip through 192.168.1.238 on smarthome-test/echo
  Bmp280Check          PASS         21s (16s deploy) 24.086358832534462°C, 942.15071591098098hPa, 53.346218103659115%RH
  HomieClientCheck     PASS         63s (18s deploy) attributes, datatypes, retained flags, /set round-trip, alert/sleeping, a refused transition and re-announce all conform
  MqttReconnectCheck   PASS         48s (15s deploy) republished and stayed subscribed across outages of 3s, 20s
  TOTAL                             174s (80s deploy)
```

The capture now stops through `Stop-SmartHomeRecordedProcess`, which reports what it did — `Stopping device debug monitor (PID 58460)...` — and no warnings were raised anywhere in the run.

Same evidence as before, from the fixed code: 14812-byte `HomieClientCheck.log` with ten `command on ...` lines and one `Reconnect handler called`, and 731 PUBLISH lines in the 131579-byte `HomieClientCheck-01-broker.log`, with the whole per-generation set alongside it.

The three review-fix paths were also exercised in isolation, since a healthy run only reaches the first: the tree is killed and the log written; an already-exited launcher reports `already gone` and does not throw; and a record whose pid has been reissued is refused — `Not stopping PID 31392 (device debug monitor): that pid now belongs to 'PING'. Leaving it alone.` — with the unrelated process left running.

RoomSensor redeployed afterwards.

PR #45, opened 2026-08-29
perf(scripts): size the deploy pad from the previous image, not a flat 400KB

### LINKED ISSUE #17: Deploy pads every image to a flat 400KB regardless of size
`Deploy-ToDevice.ps1` pads every deployment image to a flat 409,600 bytes regardless of
its actual size: WifiCheck is ~89KB, Bmp280Check ~132KB, HomieClientCheck ~181KB,
RoomSensor ~258KB.

The invariant that genuinely needs holding is only "cover the footprint of the
*previous* image", which across a suite run never exceeds ~258KB. So every deploy has
nanoff erase and verify more flash than necessary — WifiCheck by more than 4x.

Possible approach: record the last-flashed size per COM port in a small state file and
pad to `max(currentImage, lastFlashed)` rounded up to a 4KB sector. Identical
guarantee, roughly half the footprint.

Deliberately not done during the review: it touches the flashing path, and a mistake
there costs a bricked deploy rather than a failed test. The `/t:Rebuild` is load-bearing
and must not be touched.

**Files:** `scripts/Deploy-ToDevice.ps1`

JojoEffect:
Closes #17.

## What changed

`Deploy-ToDevice.ps1` padded every deployment image with 0xFF to a flat 409,600 bytes, so
`nanoff` erased and verified 400KB whether the image was RoomSensor's 259,168 bytes or
WifiCheck's 89,512. It now sizes the pad from a record of what it last flashed to that COM
port, and pads to `max(this image, that record)` rounded up to a 4KB sector.

`/t:Rebuild` is untouched.

## Why that is the same guarantee

I checked the firmware before choosing a scheme, because there is an appealing shortcut that
turns out to be wrong. `ContiguousBlockAssemblies` (`nf-interpreter`,
`src/CLR/Startup/CLRStartup.cpp`) walks the deployment region looking for assembly headers,
and on a header that fails its check it `continue`s to the next candidate rather than
breaking — over a stream whose `Length` is the *whole* deployment block range
(`BlockStorageStream_InitializeWithBlockStorageDevice`). So padding an image with one blank
sector, or writing a terminator after it, does not stop the scan: stale assemblies anywhere
in the partition are still found and loaded. That is exactly the failure the existing comment
in the script describes seeing.

Only overwriting the previous image's bytes keeps them out, which makes the invariant exactly
"cover the footprint of the previous image" — precisely what the issue proposed. The flat
409,600 held that invariant by brute force; the record holds it exactly.

The reasoning that it is sufficient: after a completed flash, everything from the image's end
up to the padded length is 0xFF, so the *next* flash only has to cover the previous image's
own length, not the previous padding. Anything beyond that has never been written by this
script.

## Conservative in every failure direction

- The record is written **pessimistically before** the flash, at the full padded length, so an
  interrupted or failed `nanoff` (or a killed shell) still leaves a record covering everything
  it might have put down. It is tightened to the image's own length only after `nanoff`
  reports success.
- Every reason the record cannot be vouched for — missing, unparseable, wrong schema version,
  wrong COM port, a different `-DeployAddress`, or the new `-FullPad` switch — falls back to
  the flat 409,600, **never** to the bare image size.
- A failure to write the pre-flash record aborts before flashing (`$ErrorActionPreference` is
  `Stop`), rather than flashing against a stale, smaller number.
- New `-DeployPartitionSize` bound replaces the old "image is larger than `-PaddedImageSize`"
  error: an image bigger than the fallback is now simply padded to itself instead of being
  rejected, which matters as device images grow (RainwaterCistern is 197,140).

## Out-of-band writers

The record only describes what this script flashed. Two other things write the deployment
partition and erase only their own footprint (`DeploymentExecuteFull` in `nf-debugger`'s
`WireProtocol/Engine.cs` erases exactly the assemblies' combined length):

- The **nanoFramework test adapter**, via `Run-Tests.ps1`. That script now clears the record
  when its run settings target real hardware, so the next deploy pads the full fallback.
- **Visual Studio's F5 deploy**, which cannot be hooked. `-FullPad` is the escape hatch, and
  it is documented in `CLAUDE.md`, the `smarthome-deploy` skill and the `device-ops` agent.

## Why a separate helper rather than a new `Get-SmartHomeDevEnvPath -Kind`

`Get-SmartHomeDevEnvPath`'s `-Port` is an MQTT port, and its `Snapshot` kind is precedent for
non-dev-env files living in that vocabulary. But this file is keyed by **COM** port, and one
parameter meaning two different kinds of port depending on `-Kind` is an ambiguity every
caller would have to hold. The new `Get-SmartHomeDeployStatePath` shares the temp directory
and the prefix (`$SmartHomeDevEnvPrefix` renamed to `$SmartHomeTempFilePrefix` accordingly),
not the signature.

## Verified on hardware

All of the following ran on the real ESP32 on COM3.

| # | Run | Pad | Result |
|---|---|---|---|
| 1 | `Deploy-ToDevice.ps1` (RoomSensor, 259,168) with no record | 409,600 (flat fallback) | Erased `0x1e0000..0x243fff`, hash verified. Record written as `StaleBytes: 259168` |
| 2 | `Deploy-ToDevice.ps1 -Project src\integrationTests\WifiCheck\WifiCheck.nfproj` (89,512) | 262,144 (covers RoomSensor) | Erased `0x1e0000..0x21ffff`. Boot capture: `Total: (10108 RAM - 89512 ROM - 44889 METADATA)` — WifiCheck's own image exactly, no leftover RoomSensor assemblies — and `[ITEST] WifiCheck PASS` |
| 3 | Same deploy again, now against an 89,512 record | 90,112 (the tight case) | Erased `0x1e0000..0x1f5fff` — 4.5x less than the old flat pad. Boot capture identical: ROM 89,512, `[ITEST] WifiCheck PASS` |
| 4 | `Run-IntegrationTests.ps1` (full suite) | 90,112 → 143,360 → 143,360 → 184,320 → 184,320 | **All 5 passed** (WifiCheck, MqttCheck, Bmp280Check, HomieClientCheck, MqttReconnectCheck) |
| 5 | `Deploy-ToDevice.ps1` (RoomSensor) to put the device back to its real job | 262,144 (image-driven) | Erased `0x1e0000..0x21ffff`, hash verified |

Run 2 is the case the padding exists for, and the ROM total is the decisive evidence: 89,512,
not 89,512 plus a tail of RoomSensor. The suite exercised the same smaller-after-larger
direction twice more — Bmp280Check (132,384) padded up to 143,360 to cover MqttCheck's
141,912, and MqttReconnectCheck (150,980) padded up to 184,320 to cover HomieClientCheck's
181,812.

**Footprint across the suite: 745,472 bytes against 2,048,000 flat — 64% less**, and the total
deploy time in the summary was 74s of the 168s run.

## Verified off hardware

A throwaway harness (not committed — this repo's unit suite runs device-side under
nanoFramework, which is the wrong home for host PowerShell helpers) asserted 41 properties of
the new `Common.ps1` functions and the padding arithmetic: round-trip; rejection of corrupt
JSON, wrong version, wrong port, missing address, missing/negative/non-numeric `StaleBytes`, a
JSON array and an empty file; per-port and all-ports clear; that a `\.\COM10`-style port
cannot escape `%TEMP%`; that the pad is always sector-aligned, always covers the image, and
always covers the previous image across a suite-like sequence of sizes. All 41 passed.

Both modified scripts and `Run-IntegrationTests.ps1` parse clean.

### DIFF
--- a/.claude/agents/device-ops.md
+++ b/.claude/agents/device-ops.md
@@ -20,8 +20,11 @@ Scripts you own (each has a matching `.claude/skills/smarthome-*` skill too):
 - Don't rewrite `Start-DevEnv.ps1`'s child launches to use `-RedirectStandardOutput`: that forces
   handle inheritance and makes piping the script's output hang forever. The ShellExecute +
   `log_dest file` + `cmd.exe` wrapper arrangement is deliberate and commented in the script.
-- `scripts\Deploy-ToDevice.ps1 [-Project <path>] [-Configuration Debug|Release] [-DeployAddress <hex>]`
-  — build + flash. `-DeployAddress` defaults to this device's current real `deploy` partition
+- `scripts\Deploy-ToDevice.ps1 [-Project <path>] [-Configuration Debug|Release] [-DeployAddress <hex>] [-FullPad]`
+  — build + flash. `-FullPad` pads the flat 400KB worst case instead of sizing the 0xFF pad
+  from the per-COM-port record of the last flash — use it on the first deploy after someone
+  has deployed from Visual Studio, which writes the deployment partition behind this script's
+  back. `-DeployAddress` defaults to this device's current real `deploy` partition
   offset (`0x1E0000`) — `nanoff`'s own hardcoded default (`0x1B0000`) landed inside the
   **factory** partition instead and silently produced a deploy the CLR could never load. If a
   deploy "succeeds" but the device never does anything afterward, verify with
--- a/.claude/skills/smarthome-deploy/SKILL.md
+++ b/.claude/skills/smarthome-deploy/SKILL.md
@@ -13,7 +13,15 @@ hard way), then flashes via `nanoff`.
 .\scripts\Deploy-ToDevice.ps1                                                    # RoomSensor, Debug
 .\scripts\Deploy-ToDevice.ps1 -Project src\devices\IrrigationControl\IrrigationControl.nfproj
 .\scripts\Deploy-ToDevice.ps1 -Configuration Release
+.\scripts\Deploy-ToDevice.ps1 -FullPad                                           # after a VS deploy
 ```
+
+The image is padded with 0xFF far enough to erase whatever the previous deploy left behind —
+sized from a per-COM-port record of the last flash, falling back to a flat 400KB whenever that
+record can't be trusted. **Pass `-FullPad` on the first deploy after anyone has deployed from
+Visual Studio**, since VS writes the deployment partition without going through this script and
+the record then describes an image that is no longer on the device. `Run-Tests.ps1` clears the
+record itself, so no `-FullPad` is needed after a unit-test run.
 
 To deploy an integration test (`src\integrationTests\*`), prefer
 `smarthome-integration-tests` — it deploys, captures, and reads the verdict in one call instead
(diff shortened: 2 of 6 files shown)


JojoEffect:
Retracting the "unrelated observation" that was in the original description (now removed from it).

I reported that `Watch-DeviceDebugOutput.ps1 -Until '\[ITEST\]'` failed to match a line that was plainly in its output, and suggested a pre-existing bug in the tool's early-exit matching. That was my mistake, not a defect.

`--until` is a plain substring match — `seen.ToString().Contains(until)` in `tools/DeviceDebugMonitor/Program.cs` — and is documented as such both there and on the script's own `-Until` parameter ("stop as soon as a line containing this text arrives"). I passed a regex, and the literal string `\[ITEST\]` is indeed not present in `[ITEST] WifiCheck PASS: ...`. `-Until '[ITEST]'` would have matched.

The suite run in this PR corroborates that the early exit works: WifiCheck's capture finished in 5s against a 75s `CaptureSeconds` budget, and Bmp280Check's in 5s against 45s. Nothing to fix, and no issue filed.

Issue #36, opened 2026-08-29
Conformance check: three known gaps left unfixed by the #32 review


JojoEffect:
Three findings from the `xhigh` code review of #32 were reported but deliberately not fixed on
branch `claude/conformance-check-review-fixes`. Each is real; each was skipped because the
available fix would have traded something away. They are recorded here so the branch can be
merged without the reasoning being lost.

The eight findings that *were* fixed are on that branch and verified: the full integration suite
passed 5/5 on hardware (ESP32 on COM3, local Mosquitto) with the conformance phase snapshot
counts identical to the pre-fix baseline.

## 1. A retransmitted duplicate can flip the settled lifecycle read

`Invoke-HomieConformanceCheck`'s refused-transition step reads `$lifecycleAfter` from
`ConvertTo-HomieSnapshot`, which collapses the capture to one entry per topic. That collapse only
merges when a payload equals the one already stored; a duplicate of an *older* payload replaces
it. So the wire order `[sleeping, alert, sleeping-dup]` leaves `Payload = 'sleeping'`, and the
settled assertion fires while all three wire assertions pass.

M2Mqtt retransmits an unacknowledged QoS-1 publish every `MqttSettings.DelayOnRetry` (1000 ms),
up to `MqttSettings.MaximumAttemptsRetry` (3) — the same retransmission #32 documents having
observed on this hardware.

Not fixed because this is arguably a *true* positive at the broker: Mosquitto re-processes a DUP
PUBLISH rather than discarding it, so the retained store really would end up holding `sleeping`.
Suppressing the assertion would hide a genuine contradiction in the retained store. What is
actually needed is a decision about which of the two is being asserted — the store's final
contents, or the device's behaviour — and they are not the same claim.

## 2. The capture tail depends on a stdio flush that `taskkill /F` does not wait for

`Stop-HomieCapture` ends the window with `taskkill /PID .. /T /F`. `mosquitto_sub`'s stdout is
redirected to a file, so the CRT uses full buffering (~4 KB) rather than line buffering, and
anything still in that buffer dies with the process.

The refused-transition step is the only caller that depends on the *tail* of a capture: the
reflected `sleeping` and the corrected `alert` are the last few dozen bytes written. It works
today only because the retained replay of ~53 topics pushes past the buffer boundary first. A
device with fewer topics, or a shorter replay, would lose the tail and the step would report
"the command was lost" against a healthy device.

[...]

PR #51, opened 2026-08-30
fix(tests): close eight review findings in the Homie conformance check

JojoEffect:
Closes eight of the eleven findings from the `xhigh` code review of #32, in the Homie v4
conformance check and the device it measures. The three left unfixed are #36 — each was
skipped because the available fix traded something away, and that reasoning lives on the
issue rather than being lost when this merges.

Two files: `scripts/Run-IntegrationTests.ps1` and
`src/integrationTests/HomieClientCheck/Program.cs`.

## The four that could fail a healthy device

- **The refused-transition step searched the whole capture window for the correction.** The
  window is deliberately opened *before* the command, so it also holds payloads that predate
  it — the preceding step republishes its command every poll round, and M2Mqtt retransmits an
  unacknowledged QoS-1 publish every `DelayOnRetry` (1s), three times over. A first-occurrence
  `IndexOf` could find one of those and report a correct device as having published the
  correction too early. The correction is now looked for *after* the reflection, with a
  separate unordered search kept only to tell "never corrected" from "corrected too early".
- **`Start-HomieCapture`'s `-WaitForConnectSeconds` could be defeated by the file it was
  waiting on.** It takes "the output file has bytes" as proof its own subscriber is live. The
  previous capture is torn down with `taskkill /F`, which returns before Windows releases the
  handle, so `Remove-Item` could fail with a sharing violation — silenced by
  `-ErrorAction SilentlyContinue` — leaving the previous capture's bytes to satisfy the guard
  instantly. That is the exact race the parameter was added to prevent. Removal is now proved,
  or the run fails saying so.
- **The refused step was not bracketed by `try`/`finally`.** A throw from the publish left a
  live `mosquitto_sub` appending to the one snapshot path every later capture reuses, quietly
  invalidating the retain flags for the rest of the suite.
- **`ConvertTo-HomieSnapshot`'s "same payload" merge used `-eq`,** which is case-insensitive in
  PowerShell, so a live `TRUE` merging with a replayed `true` inherited a retain flag it never
  earned — the very inheritance the replace-on-difference branch exists to prevent. Now `-ceq`.

## The rest

- An empty capture window reported the forbidden transition as *applied*, blaming the device
  for a host-side failure. It now says nothing was measured, as the wire assertions already did
  for a lost command.
- `Write-ConformancePhaseBreakdown` ran only on the check's return paths, so a throw skipped
  the timings for the run that needed them most. Now in a `finally`.
- The `'%t %r %p'` line regex had a second copy in the refused step, right next to a comment
  explaining why the subscriber arguments are *not* respelled there. One reader now:
  `ConvertFrom-HomieCaptureLine`.
- `HandleLifecycleCommand`'s switch defaulted to `Ready()`, so a state added to
  `LifecycleStates` — which is also what builds `$format`, and so what controllers are offered
  — would silently apply the wrong transition. Every arm is named and the default refuses.

## Rebase: what survived unchanged, and what had to be re-expressed

The branch was cut before `e32d936` and `2d49328` and has been rebased onto `115005e`.
`e32d936` added managed-debug-output capture and broker-log preservation to the same
conformance path this branch expands, so the two met at one call site.

**Seven of the eight findings survived the rebase unchanged** — the `Start-HomieCapture`
removal proof, the refused-step `try`/`finally`, the `-ceq` merge, the ordered correction
search, the empty-window report, `ConvertFrom-HomieCaptureLine`, and the `Program.cs` switch
all applied to regions main had not touched.

**One had to be re-expressed: the `Write-ConformancePhaseBreakdown` finally.** Main had
independently introduced a `try`/`finally` at that call site — but for stopping the debug
capture, with the breakdown left *outside* it. The finding was therefore still live on main
in the same form, and the fix here joins main's existing `finally` rather than nesting a
second one inside it. Ordering is deliberate: after `Stop-DeviceDebugCapture`, because that
call releases the COM port the next deploy needs, and `2d49328` made it warn rather than throw
on a recycled or already-dead pid — so going second costs the breakdown nothing. The run log
confirms the order on the wire: `Stopping device debug monitor (PID 35260)...` immediately
followed by `phase breakdown`.

Nothing from `e32d936`/`2d49328` was dropped or duplicated: the run preserved
`HomieClientCheck.log` (20 KB of managed debug output), plus the broker and `homie/#` logs.

## Verification

**Run on hardware** — the original 5/5 was against the pre-rebase base and proved nothing
about this result, so the whole suite was re-run after the rebase.

Full integration suite on ESP32 / COM3 with local Mosquitto, **5/5 PASS**, 171s total (79s of
it deploys):

| Test | Result |
|---|---|
| WifiCheck | PASS 22s |
| MqttCheck | PASS 20s |
| Bmp280Check | PASS 19s |
| HomieClientCheck | PASS 62s |
| MqttReconnectCheck | PASS 48s |

Conformance phase snapshot counts **3/0/1/5/2**, identical to the pre-fix baseline — so none of
this changed what the check measures. The check itself took 45.1s against a documented
43.1–44.1s; the drift is `e32d936`'s debug-capture launch, not anything in this branch.

The refused transition is visible in the preserved `homie/#` log as `sleeping` then `alert` on
`matrix/lifecycle`, in that order, against the device's own `'sleeping' -> False` — which is
precisely what the reordered search asserts.

Also run: `msbuild SmartHome.sln /t:Build` clean across all 14 projects (warnings all
pre-existing). RoomSensor was redeployed afterwards, so the device is back to its real job
rather than holding the last integration test.

**Not run:** the unit suite (`Run-Tests.ps1`) — CI runs it on the nanoclr virtual device, and
this branch touches no unit-tested code.

## Follow-up: two findings from the review of this branch

`2af2b12`, in the same two files. Both are in code this branch introduced.

- **The refused step looked for the correction twice, and the second search could not
  support the claim it made.** Alongside the ordered search described above, an unordered
  `IndexOf` across the whole window existed only to tell "never corrected at all" from
  "corrected before the reflection it was supposed to overwrite". But it is the same
  first-occurrence search the ordered one replaced, with the same exposure: the window opens
  before the command, so it still holds the preceding step's live traffic. A device that
  never published the correction was therefore reported as having published it too early —
  sending the next reader after an ordering bug that did not happen, and hiding the one that
  did. The two branches are now one message carrying the observed payload sequence, so a
  genuinely early correction stays visible as evidence rather than as a claim the runner
  cannot support.
- **`HandleLifecycleCommand`'s new default arm returned a bare `false`,** which reaches the
  same `'{payload}' -> {applied}` line as a transition the convention's own state machine
  refused — the "no sign that anything was missed" the named arms were added to remove. It
  now goes through `RefuseUnwiredState`, which says the state is offered in `$format` but has
  no transition wired.

**Verification** — the whole suite re-run on hardware after the fixes: ESP32 / COM3 with local
Mosquitto, **5/5 PASS**, 172s (77s of it deploys). WifiCheck 21s, MqttCheck 20s, Bmp280Check
19s, HomieClientCheck 62s, MqttReconnectCheck 50s. The refused transition is in the preserved
`homie/#` log as `sleeping` then `alert` on `matrix/lifecycle`, in that order, which is what the
ordered search asserts. `HomieClientCheck.nfproj` also builds clean on its own (warnings all
pre-existing).

**Correction to the Verification section above:** it says RoomSensor was redeployed afterwards.
That was true of the earlier run, but this re-run left `MqttReconnectCheck` on the device, so
RoomSensor needs redeploying before the device is back to its real job.

## #36

Still accurate against the rebased code, and the line numbers it cites are unaffected — it
names functions and behaviours, not line numbers. Item 3 (every accepted lifecycle command
publishing twice) is directly visible in this run's `homie/#` log as the duplicate pairs on
`matrix/lifecycle`.

### DIFF
--- a/scripts/Run-IntegrationTests.ps1
+++ b/scripts/Run-IntegrationTests.ps1
@@ -699,7 +699,26 @@ function Start-HomieCapture {
     )
 
     $out = Get-SmartHomeDevEnvPath -Port $Port -Kind Snapshot
-    Remove-Item -Path $out -Force -ErrorAction SilentlyContinue
+
+    # Removed, and *proved* removed. The previous capture is torn down with taskkill /F,
+    # which returns before Windows has released cmd.exe's handle on this file, so
+    # Remove-Item can fail with a sharing violation -- and -ErrorAction SilentlyContinue
+    # would swallow that. A surviving file is not merely untidy: the connect wait below
+    # takes "the file has bytes" as proof that THIS subscriber is live, and the previous
+    # capture's bytes satisfy it instantly, defeating the wait entirely.
+    $removeDeadline = (Get-Date).AddSeconds(5)
+    while (Test-Path -Path $out) {
+        Remove-Item -Path $out -Force -ErrorAction SilentlyContinue
+        if (-not (Test-Path -Path $out)) {
+            break
+        }
+
+        if ((Get-Date) -ge $removeDeadline) {
+            throw "Could not clear the snapshot capture file '$out': a previous subscriber still holds it open."
+        }
+
+        Start-Sleep -Milliseconds 50
+    }
 
     $sub = Get-MosquittoTool -Name 'mosquitto_sub.exe'
 
@@ -747,9 +766,40 @@ function Stop-HomieCapture {
     $script:snapshotsTaken++
 
     Start-Sleep -Seconds $SettleSeconds
+    # See #36: taskkill /F does not wait for mosquitto_sub to flush. Its stdout is
+    # redirected to a file, so the CRT buffers fully (~4KB) rather than by line, and
+    # whatever is still in that buffer dies with the process. Only the refused-transition
+    # step depends on the TAIL of a window -- every other caller needs the bulk retained
+    # replay, which is far past the buffer boundary by the time the window closes.
     Stop-SmartHomeProcessTree -ProcessId $Capture.ProcessId
 
     return @(Get-Content -Path $Capture.Path -ErrorAction SilentlyContinue)
+}
+
+function ConvertFrom-HomieCaptureLine {
+    # The one reader of the '%t %r %p' line layout, for the same reason Start-HomieCapture
+    # takes the subscriber arguments from Common.ps1 rather than respelling them: two
+    # callers now read these lines -- the per-topic collapse below and the refused step's
+    # ordered read -- and a second copy of the regex is a place where a change to the
+    # format could be fixed in one reader and silently keep parsing in the other.
+    #
+    # Returns $null for a line that is not a message: mosquitto_sub's stderr shares the
+    # capture file.
+    param(
+        [string]$Line
+    )
+
+    # "<topic> <0|1> <payload>", and the payload may itself contain spaces.
+    $match = [regex]::Match($Line, '^(\S+)\s+([01])\s?(.*)$')
+    if (-not $match.Success) {
+        return $null
+    }
+
+    return @{
+        Topic    = $match.Groups[1].Value
+        Retained = $match.Groups[2].Value -eq '1'
+        Payload  = $match.Groups[3].Value
+    }
 }
 
 function ConvertTo-HomieSnapshot {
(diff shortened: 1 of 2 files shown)


JojoEffect:
Filed #54 for something adjacent that this PR makes sharper.

`Stop-HomieCapture` still identifies its subscriber by a bare pid and kills it with `Stop-SmartHomeProcessTree`, which throws a `NativeCommandError` when the process has already exited — the same hazard `2d49328` fixed for the managed-debug capture, applied there but not here. `mosquitto_sub` exits within ~3s when it cannot reach the broker, and the settle before the kill is exactly 3s, so an unreachable broker hits it as a matter of course rather than as a race.

Relevant to this PR specifically: the `try`/`finally` added around the refused-transition step means a throw from `Stop-HomieCapture` now replaces the original exception. The `finally` is still correct — closing the window on a failed publish is the point — but it turns a misleading error message into a masked verdict, which is worth knowing while reviewing. #54 has the measurements and a suggested fix; it is deliberately out of scope here.

PR #55, opened 2026-08-30
feat(scripts): add a git cleanup script for merged worktrees and branches

JojoEffect:
## What

Adds `scripts\Clean-GitBranches.ps1` and the matching `smarthome-clean-branches` skill: one call that classifies every worktree and every local and remote branch against `origin/main` and, on request, removes the merged ones in a single batch.

Nothing in the repo said which branches were finished. The answer was recoverable, but only by hand — `git branch --merged`, `git worktree list`, `gh pr list` and a `rev-list` per branch, re-derived from scratch every time anyone asked. This repo had accumulated 34 local branches, 29 of them fully contained in `origin/main`. The script does that analysis once, so a future session spends one script call instead of a dozen exploratory git invocations.

Report-only unless `-Delete` is passed. It needs no `local.env.ps1`, no restored packages and no device, so it runs in a fresh worktree with no setup.

## What counts as safe to delete

A branch is a candidate only when its work is provably already in the base branch:

- **Ancestry** — the tip is an ancestor of `origin/main`. Same test git uses to allow `git branch -d`, and the only one that proves no commit is lost. Deleted with `-d`.
- **Squash/rebase merge** — the tip is *not* an ancestor, but `gh` reports the pull request as MERGED. GitHub rewrites the commits, so ancestry cannot see them. Own group, deleted with `-D`, because there the evidence is GitHub's word rather than local history.

Kept, with the reason printed per entry: the base branch, anything a staying worktree pins, anything with an open pull request, anything unmerged (with its ahead-count), anything in `-Protect`. An open pull request outranks a merged one on the same branch name, so a reused name protects the branch rather than freeing it.

## Worktrees

Always analysed and reported; `-Worktrees` is what allows removal. Every state is handled explicitly:

| State | Handling |
|---|---|
| directory gone, admin record left | prune |
| main worktree | keep — git refuses, and it is the repository |
| the one the script runs in | keep — removing it deletes the session's own cwd |
| locked | keep, reason shown |
| uncommitted or untracked changes | **keep, always** |
| pins a protected branch or one with an open pull request | keep |
| clean, HEAD already in `origin/main` | remove |
| clean, carries commits `origin/main` lacks | keep, count shown |

A dirty worktree is never removed and there is deliberately no flag to override that. Uncommitted work is the one thing neither the reflog nor the remote can give back; every other mistake this script could make is recoverable. `git worktree remove` is called without `--force` for the same reason — the dirty check already excluded everything that would need it, so a refusal means the state changed underneath and is worth surfacing.

With `-Worktrees -Delete` the two passes chain: worktrees are removed first, then the branches they were pinning are deleted in the same run. Branch classification runs against the post-removal worktree set, so the report shows those branches as candidates (`freed by removing worktree <name>`) up front rather than describing a state the run is about to invalidate.

## Two PowerShell 5.1 traps, pinned by comments where they bite

Both found by the script misbehaving, not by reading:

- `ConvertFrom-Json` emits a JSON array as **one** object, so `@()` wraps it instead of unrolling it. The pull-request loop ran once with the whole array bound to `$pr`, and `$pr.isCrossRepository` — an array of booleans — was truthy, silently skipping every pull request and disabling open-pull-request protection.
- `%(refname:short)` shortens `refs/remotes/<remote>/HEAD` to the bare remote name, which no longer carries the prefix the branch name is cut from.

## Common.ps1

The memo initialisation now passes `-WhatIf:$false`. It is dot-sourced by this script, the repo's first with `SupportsShouldProcess`, and `$WhatIfPreference` propagates into it — so a `-WhatIf` run reported `WhatIf: Set variable SmartHomeMSBuildPath` as though it were an action the user had asked about, in front of the real preview.

## Verification

No hardware involved — this script touches no COM port, opens no broker and builds nothing, so the usual hardware caveat does not apply here. Exercised against this repository:

- report (default), `-Scope Local`, `-Scope Remote`, `-Json`
- `-Worktrees` report: 4 removable worktrees found, and the branch one of them pinned correctly re-classified as a delete candidate
- `-WhatIf`: previews 4 worktree removals plus the freed branch, changes nothing (7 worktrees, 5 branches before and after)
- the real batch delete: 29 local branches removed, `main` / the current branch / 3 worktree-pinned branches / the open-PR branch all kept, remote untouched because it had no candidates
- dirty-worktree guard: one untracked file added to a removal candidate flips it to KEEP

CI builds and runs the unit tests as usual; neither covers a PowerShell dev script.


JojoEffect:
Ran the cleanup for real after opening this, and it turned up a defect in the worktree pass — fixed in d83360b.

Three of four worktrees removed cleanly. The fourth failed with `Permission denied`, and the warning claimed *"the branch it pins will be reported as still pinned"*. That was wrong. `git worktree remove` is not atomic: it unregisters the worktree and *then* deletes the directory, so on Windows a process holding something in there fails the second half after the record is already gone. The branch was in fact freed and correctly deleted, and an orphaned directory was left on disk that the message never mentioned.

The removal pass now asks git which of the two states it ended in:

- **still registered** — the branch stays pinned, so it is dropped from the delete batch and one failure produces one warning instead of two
- **unregistered, directory left** — an orphan nothing points at. The branch delete is correct and proceeds; the warning prints the exact `Remove-Item -Recurse -Force '<path>'`. This case cannot be fixed by re-running, because `git worktree list` no longer mentions the directory.

Both paths reproduced against a scratch worktree: holding its `.git` pointer file open gives the orphan case, additionally holding the admin `gitdir` file open gives the still-registered case. Branch deleted in the first, correctly left alone in the second, exit 1 in both.

Worth noting for anyone reviewing the design: the process that blocked the delete was holding an *empty* directory — a shell whose cwd is inside a worktree locks it on Windows without any file being open. So "clean" really does not imply "idle", and `git worktree lock` remains the way to mark one as in use.

Issue #74, opened 2026-09-01
Host-side script logic has no test home, so every fix to it needs a throwaway harness


JojoEffect:
The `scripts\` directory carries a large amount of logic that decides test verdicts, and none of it has a test home. Proving a change to it means either running the whole on-device integration suite, or building a throwaway harness that is deleted afterwards.

## What prompted this

While fixing #35, two new host-side functions were added to `Run-IntegrationTests.ps1` — `Wait-ForAnnounceWitnessed` and `Get-SubscriberLogLineCount`. Both are pure log-reading with no device, no broker and no network, so they are exactly the kind of thing that can be proved at a desk in under a second.

Proving them meant writing a harness that extracts the function definitions out of the shipped script via the PowerShell AST, dot-sources them into the test scope, stubs `Get-SmartHomeDevEnvPath`, and runs them against fabricated `homie/#` logs. It found nothing wrong, and it exercised eight cases including the decisive one — that a previous boot's `$state=init` sitting before the watermark must not satisfy the wait, which is the regression the whole fix exists to prevent.

That harness lives in a temp directory and is not in the repository, so the next change to those functions has no way to re-run it. The same is true of the harness written for #59, which measured `Start-HomieCapture`'s connect wait returning after 2.39s of a 5s budget, and of the measurements in #36/#65 that refuted the stdio-buffering hypothesis. Each was built from scratch, used once, and thrown away.

## Why it matters here specifically

This repository's own history is the argument. `CLAUDE.md` records a green unit run that executed nothing for three commits after the `NFUnitTest` rename, and the conformance check has had two separate defects where a check passed while lying. The host-side half of the suite is where the verdict is actually decided — `Invoke-HomieConformanceCheck`, `Invoke-BrokerOutageCheck`, `Wait-ForRetainedValue`, `Start-HomieCapture`/`Stop-HomieCapture`, `ConvertFrom-HomieCaptureLine` — and it is the half with no automated coverage at all.

It is also the half CI could actually run. #22 covers building the solution and running the nanoFramework unit tests on the virtual device; those tests cannot reach a PowerShell function. A host-side suite would be the first thing in this repository that CI can run against the integration tooling itself.

## What closing this would look like

Roughly in order of cost:

[...]

PR #75, opened 2026-09-01
fix(integration): witness this boot's announce before trusting a retained $state

JojoEffect:
Closes the root cause behind #35: the conformance check could accept the *previous* instance's retained `$state=ready` as proof that the device it just flashed had finished announcing.

## What was wrong

`Wait-ForRetainedValue` required the retain flag and treated a retained `$state=ready` as the "device has finished announcing" signal. Its own comment said so:

> `$state` arriving retained is, in effect, the "device has finished announcing" signal, and this turned it from an assumed invariant into an enforced one.

That does not hold. Retained state outlives the connection that published it, and the will that would replace it fires on keepalive expiry rather than on the reset a deploy causes. So while the old instance is still registered at the broker, its `$state=ready` sits in the store and any snapshot taken in that window is handed it.

The consequence is #35's symptom exactly. The host skips the wait it should have spent, reaches the `/set` phase roughly 7s early, and publishes five non-retained `/set` commands before the new instance has subscribed at the broker. A non-retained publish with no matching subscription is dropped with no trace, and the five properties are still holding their boot values when the check reads them back. The failing run's signature matches: one announce snapshot in 3.4s where a healthy run takes three in ~10s.

## What changed

The announce is now witnessed live before any retained `$state` is believed.

`HomieClient.Announce()` publishes `$state=init` before anything else and only transitions to the post-init state once the device info is out, and `Connect()` calls `Announce(State.Ready)` — so an `init` is published exactly once per announce, by the instance doing it. That is the per-boot signal a retained `ready` cannot give.

Three details worth knowing:

- **It is read from the long-running `homie/#` log, not from a snapshot.** `init` is transient — the device is in it for milliseconds and the next state overwrites it in the retained store — so a 3s snapshot would usually miss it. The detached subscriber is connected throughout and records every message as it arrives.
- **The watermark is taken before the deploy**, in the run loop, not inside the verdict function. By the time a verdict function runs, the device has already been reset and may already have announced. Taking it earlier also means an announce that lands while the deploy is still finishing still counts, so there is no race to lose.
- **Both waits share the one `SettleSeconds` budget**, with a floor of one snapshot for the `ready` wait, so a healthy run cannot take longer than it did before and a slow boot cannot be failed by a zero-second remainder.

The stale comment in `Wait-ForRetainedValue` is replaced with what the retain flag does and does not establish.

## Verification

**On hardware** — `Run-IntegrationTests.ps1 -Tests HomieClientCheck`, real ESP32 on COM3, real broker:

```
Waiting up to 90s for homie-client-check to announce...
  announce witnessed live ($state=init).
  retained store holds 53 topics
...
  phase breakdown (3s per snapshot):
    announce                  9.2s   2 snapshot(s)
    attributes                0.0s   0 snapshot(s)
    /set round-trip           3.6s   1 snapshot(s)
    out-of-format /set        3.8s   1 snapshot(s)
    lifecycle                17.4s   5 snapshot(s)
    re-announce              13.7s   2 snapshot(s)
HomieClientCheck: PASS
```

67s total, 19s of it deploy. The announce phase is 9.2s against the 9.9–10.3s the passing runs in #35 recorded, and the `/set` round-trip took one snapshot, so no command needed #32's retry.

**At a desk** — the two new functions are pure log reading, so they were exercised against fabricated `homie/#` logs by extracting them from this script via the PowerShell AST. Eight cases, all passing, including the decisive one:

```
PASS  old boot's init before the watermark is not accepted
PASS  this boot's init after the watermark is accepted
PASS  'init' on another topic is not an announce
PASS  a longer device id sharing this prefix is not this device
PASS  a retained-flagged init still counts
PASS  a missing log counts 0 rather than throwing
PASS  and the wait then returns false rather than throwing
PASS  a one-line log counts 1
```

That harness is not in this diff, because the repository has nowhere to put it — filed as #74, which is also where the argument for a host-side test home lives.

## What this does not do

It does not prove that the retained-`ready` path is what caused the observed 1-in-8 failure on 2026-08-26. That run left almost no evidence, which is what the first half of #35 was about and what earlier work has since fixed. What it does is remove a mechanism that produces exactly that symptom and that the code was documented as relying on. The phase breakdown remains the tell if it recurs: a `/set` phase costing more than one snapshot means a command was lost and retried.

Refs #35

### DIFF
--- a/scripts/Run-IntegrationTests.ps1
+++ b/scripts/Run-IntegrationTests.ps1
@@ -704,8 +704,9 @@ function Wait-ForAnnounceWitnessed {
     # would usually miss it entirely. The detached subscriber is connected throughout and
     # records every message as it arrives.
     #
-    # Read-only and race-free: the watermark is taken before the deploy, so an announce that
-    # lands while the deploy is still finishing is still counted. Nothing is published here.
+    # The watermark is what makes this per-boot rather than per-device: it is read in the
+    # gap between the flash's hard reset and the new image's first publish, so a line after
+    # it cannot have come from the image that was replaced. Nothing is published here.
     param(
         [string]$Port,
         [string]$DeviceId,
@@ -792,16 +793,17 @@ $script:snapshotsTaken = 0
 # conformance check ever assigns it.
 $script:currentPhase = $null
 
-# How many lines the long-running homie/# log held immediately before this test's deploy.
-# Everything after it is traffic from the image this test just flashed; everything before
-# it belongs to whatever was running on the device beforehand -- which, when the previous
-# suite run left HomieClientCheck flashed, is the *same* device id announcing on the same
-# broker. Wait-ForAnnounceWitnessed reads from here down, so an announce it witnesses is
+# How many lines the long-running homie/# log held the moment this test's flash finished.
+# Everything after it is traffic from the image just flashed; everything before it belongs
+# to whatever was running on the device beforehand -- which, when the previous suite run
+# left HomieClientCheck flashed, is the *same* device id announcing on the same broker.
+# Wait-ForAnnounceWitnessed reads from here down, so an announce it witnesses is
 # necessarily this boot's.
 #
-# Taken in the run loop rather than inside a verdict function because it has to be read
-# before the deploy, and a verdict function only runs after one. Declared here for the
-# same StrictMode reason as the counter above.
+# Taken in the run loop rather than inside a verdict function, so that it is read in the
+# gap between the flash's hard reset and the new image's first publish -- see the comment
+# at the read for why neither side of that gap will do. Declared here for the same
+# StrictMode reason as the counter above.
 $script:subscriberLogWatermark = 0
 
 function Get-HomieRetainedSnapshot {
@@ -1383,9 +1385,11 @@ function Measure-HomieConformance {
     # and published five non-retained commands into a window the new instance had not yet
     # subscribed in (#35).
     #
-    # Both waits share the one SettleSeconds budget, so this cannot make a healthy run take
-    # longer than it did before; on a healthy run the init lands within a second of boot and
-    # costs almost nothing.
+    # Both waits come out of the one SettleSeconds budget: the witness may take all of it,
+    # and whatever is left goes to the retained-ready wait below. On a healthy run the init
+    # lands within a second of the device joining, so the witness costs almost nothing and
+    # the ready wait keeps the budget it had before this check existed. The floor below is
+    # the one way the phase can run past SettleSeconds, and it costs at most one snapshot.
     $announceDeadline = (Get-Date).AddSeconds($Settings.SettleSeconds)
     $witnessed = Wait-ForAnnounceWitnessed -Port $Port `
                                            -DeviceId $deviceId `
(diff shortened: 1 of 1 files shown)


JojoEffect:
A review pass over `892032e` found one real hole in it, now fixed in `d634a20`.

**The watermark was read before the deploy**, which leaves the whole rebuild-and-flash window — 19s on the verification run — open for the image *already on the device* to reconnect to the freshly started suite broker and re-announce. `ReconnectingMqttClient` retries, `HandleConnectionOpen` takes the device back through `init`, and that `$state=init` lands after a pre-deploy watermark. The witness would then be satisfied on behalf of the boot that is about to be erased, which is exactly the failure it exists to stop.

The trigger is not exotic: running `Run-IntegrationTests.ps1 -Tests HomieClientCheck` twice in a row is enough, and that is the loop-the-test reproduction #35 itself recommends. Run 1 leaves `HomieClientCheck` flashed and running, its broker is stopped at the end of the run, and run 2's `Start-SuiteBroker` is what the old instance reconnects to.

Reading the watermark immediately *after* the flash closes the window from both sides. `nanoff`'s last act is a hard reset, so the previous image is gone and cannot publish again, while the newly flashed one still has a boot, a WiFi association and an MQTT connect ahead of it before it can announce.

Two smaller corrections from the same pass:

- The `NO-RESULT` detail for the retained-ready wait reported `SettleSeconds`, but the wait had actually run for whatever the witness left of that budget. A slow announce would have been reported as a device given 90s when it had 5. It now names both numbers.
- The comment claiming the two waits cannot push the phase past `SettleSeconds` was not what the code does — the one-snapshot floor under the remaining budget can take it past. Stated plainly instead.

## Verification status

The desk harness still passes 8/8 against the shipped source, including the case that pins this behaviour (a previous boot's `init` before the watermark must not satisfy the wait).

**The hardware run quoted in the description predates these commits.** It exercised `892032e`, not `d634a20`, and the change moves when the watermark is read — so it is worth one more `Run-IntegrationTests.ps1 -Tests HomieClientCheck` before merging. The failure mode if the timing is wrong is loud rather than silent (the witness would time out and the run would report `NO-RESULT`), but it has not been observed either way on this commit.

JojoEffect:
Hardware verification re-run against `d634a20`, so the evidence now matches the head of the branch rather than the commit before it.

`Run-IntegrationTests.ps1 -Tests HomieClientCheck`, real ESP32 on COM3, real broker:

```
Waiting up to 90s for homie-client-check to announce...
  announce witnessed live ($state=init).
  retained store holds 72 topics
...
  phase breakdown (3s per snapshot):
    announce                  9.9s   2 snapshot(s)
    attributes                0.0s   0 snapshot(s)
    /set round-trip           3.7s   1 snapshot(s)
    out-of-format /set        4.0s   1 snapshot(s)
    lifecycle                17.4s   5 snapshot(s)
    re-announce              13.5s   2 snapshot(s)
HomieClientCheck: PASS
```

66s total, 17s of it deploy. The witness still fires with the watermark read after the flash instead of before it, which was the open question — the failure mode if the new timing had been wrong is a `NO-RESULT` from a witness that never matches, and it did not occur. The announce phase is 9.9s / 2 snapshots against 9.2s / 2 on `892032e` and the 9.9–10.3s the passing runs in #35 recorded, and the `/set` round-trip took one snapshot, so no command needed #32's retry.

One incidental confirmation: the retained store reads 72 topics here against 53 on the earlier run. The difference is `room-sensor-office`, which was redeployed between the two runs and is publishing into the same broker. That matches the reading already recorded in #35 — the snapshot count is cross-device and is not a signal about this device — and is why the count is reported rather than asserted on.

RoomSensor was redeployed afterwards, so the device is back on its real job.

Issue #85, opened 2026-09-01
Five duplications between Run-IntegrationTests.ps1 and Start-DevEnv.ps1 that #74 did not absorb


JojoEffect:
A `/simplify` pass over `scripts\` on `main` found five duplications between `Run-IntegrationTests.ps1` and `Start-DevEnv.ps1`, and inside `Run-IntegrationTests.ps1` itself. They were recorded on #74 rather than filed, on the reasoning that #74 would restructure both files and separate issues would collide with it.

#74 has now landed and did not touch them: it took the run-loop guard route (a dot-source of `Run-IntegrationTests.ps1` defines its catalog and functions and runs nothing), which made the file testable without moving any helper. So the inventory needs a home of its own. It is reproduced here from [the comment on #74](https://github.com/JojoEffect/SmartHome/issues/74#issuecomment-5498606566); all of it was confirmed by reading the source, not proposed.

## 1. `Get-MosquittoTool` is duplicated

`Run-IntegrationTests.ps1` resolves a mosquitto binary under `SMARTHOME_MOSQUITTO_DIR` and fails with ``"Not found: {0}`nCheck SMARTHOME_MOSQUITTO_DIR in local.env.ps1."``; `Start-DevEnv.ps1` does the same for `mosquitto.exe` and `mosquitto_sub.exe` with the byte-identical message. The runner's own comment already admits it: "Same guard and remediation Start-DevEnv.ps1 gives the same two binaries."

## 2. The subscriber command-line rendering is duplicated verbatim

Two lines, character for character, in both files:

```powershell
$subscriberArgs = (Get-SmartHomeSubscriberArguments -Port $Port |
    ForEach-Object { if ($_ -match '[\s/#]') { '"{0}"' -f $_ } else { $_ } }) -join ' '
```

`Get-SmartHomeSubscriberArguments` is already shared in `Common.ps1`; the `cmd.exe`-quoted *rendering* of it is not, which is the half that breaks if the argument list ever grows a value needing different quoting. The runner's comment says "Quoting idiom matches Start-DevEnv.ps1's subscriber launch", and nothing enforces that.

## 3. Three subscriber-log poll loops with one shape

[...]

Issue #84, opened 2026-09-01
Host-side tests stop at the pure functions; the verdict functions that own a process or a broker are still uncovered


JojoEffect:
`scripts\tests` covers the pure host-side functions. The ones that actually reach the broker and the device log — the ones that decide a verdict — are still untested, and one of them cannot be reached from a test at all.

## Where this comes from

#74 gave `scripts\` a test home: `Run-ScriptTests.ps1`, a runner under `scripts\tests`, a dot-source guard so `Run-IntegrationTests.ps1` can be loaded without running the suite, and a CI step. It covered what is pure — `Get-CatalogValidationError`, `ConvertFrom-HomieCaptureLine`, `ConvertTo-HomieSnapshot`, `Get-HomieLivePayloads`, `Get-SubscriberLogLineCount`, `Wait-ForAnnounceWitnessed`, `Get-ConformanceCaptureSeconds`, plus `Common.ps1`'s globs, dev-environment state and the deploy-padding record.

It deliberately stopped short of everything that owns a process, a file handle or a broker. That was to keep the diff reviewable, and it is recorded in `.claude\skills\smarthome-script-tests\SKILL.md` as the suite's stated boundary. This issue is the rest.

## What is still uncovered

- **`Start-HomieCapture` / `Stop-HomieCapture`.** The connect wait was rewritten under #59 and proved with a throwaway harness (`Test-ConnectWait.ps1`, 25 cases) that ran the pre-change and patched versions side by side against a scripted fake subscriber — a `cmd.exe` script whose stdout stands in for `mosquitto_sub`'s. That technique is the model for covering these: no broker is needed, only a process that writes the right bytes at the right times. The harness is gone; the cases are not in the repository.
- **`Wait-ForRetainedValue`.** Polls a snapshot for a topic to reach a value, and checks the retain flag. Its previous defect — treating a retained `$state=ready` from a *previous* boot as proof this one had announced — is what #35 was about, and the replacement (`Wait-ForAnnounceWitnessed`) is covered while this one is not.
- **`Invoke-BrokerOutageCheck`.** Decides `PASS` / `RESTARTED` / `FAIL` from heartbeat counters across a broker restart plus an echo round-trip. The counter comparison and the echo requirement are pure given a captured log; only the broker restart is not.
- **`Invoke-HomieConformanceCheck` / `Measure-HomieConformance`.** The largest verdict function in the repository, and the one with the worst record: two separate defects where the check passed while lying (#34, #36).
[...]

Issue #80, opened 2026-09-01
The rest of scripts\ still addresses concrete paths with -Path, so a checkout path containing [ or ] breaks deploy, test and release too


JojoEffect:
The rest of `scripts\` still addresses concrete paths with `-Path`, so a checkout path containing `[` or `]` breaks deploy, test and release too

## What is wrong

#71 fixed `Restore-Packages.ps1`, `Test-Setup.ps1` and the `Common.ps1` helpers those two reach, so a checkout at a path containing `[` or `]` can restore its packages and pass its preflight. That was the reviewable half. The same `-Path` pattern is still in the other ten scripts — about 70 call sites — and it fails the same way for the same reason: `-Path` reads `[` and `]` as wildcard character-class syntax, both are legal in a Windows directory name, and every one of these paths is a concrete directory or file derived from `$PSScriptRoot`, a config value, or a build output. None of them ever wants wildcard semantics.

#71 also noted this and left it deliberately: it would not have stayed a reviewable diff next to the fix.

## Reproduction (verified, no hardware)

Measured against a synthetic checkout at `%TEMP%\claude\sweep-probe\SmartHome [wip]`, containing `src\devices\RoomSensor\Properties\AssemblyInfo.cs` and `src\devices\RoomSensor\RoomSensor.nfproj`:

```
Set-AssemblyVersion.ps1:71   Get-ChildItem -Path        <root>\src -Recurse -Filter AssemblyInfo.cs  ->  0
                             Get-ChildItem -LiteralPath <root>\src -Recurse -Filter AssemblyInfo.cs  ->  1

Run-Tests.ps1:58             Test-Path        <root>\src\devices\RoomSensor\RoomSensor.nfproj         ->  False
Deploy-ToDevice.ps1:108      Test-Path -LiteralPath  (same file)                                      ->  True

Common.ps1:219               Get-Content -Path (same file) -Raw                                        ->  throws ItemNotFound
```

What each of those turns into:

- **`Set-AssemblyVersion.ps1:71`** — zero files, so the script takes its own `$files.Count -eq 0` branch and exits 1 with `No AssemblyInfo.cs found under src. Has the layout changed?`. `.nfproj` has no generated assembly info, so this script is the only thing that makes a release build carry its version (`CLAUDE.md`): on a bracketed checkout the release workflow's version step fails, and the message points at the repository layout rather than at the path.
[...]

Issue #79, opened 2026-09-01
packages\ is never pruned, and the test adapter is picked by sort order rather than by reference


JojoEffect:
`Restore-Packages.ps1` only ever adds to `packages\`: it copies a missing `packages\<Id>.<Version>\` in from the local NuGet cache and never removes one the checkout has stopped referencing. Nothing else prunes that folder either.

Measured on this machine today, against the main checkout:

| | count |
|---|---|
| package versions referenced by its 14 `packages.config` | 29 |
| folders present in `packages\` | 61 |
| present but not referenced | 32 |

The 32 are older versions from this repo's history (`nanoFramework.M2Mqtt.5.1.146` and `5.1.206` alongside the referenced `5.1.221`, `nanoFramework.Hardware.Esp32.1.6.19` and `1.6.37` alongside `1.6.42`, and so on) — plus, until #68 is fixed, anything a worktree referenced when a restore happened to run from the main checkout.

## Why it is more than clutter

`Get-NanoFrameworkTestAdapterDir` in `scripts\Common.ps1` resolves the unit-test adapter by taking the **highest-sorting** `nanoFramework.TestAdapter.dll` path under `packages\`:

```powershell
$adapterDll = Get-ChildItem -Path $packagesDir -Recurse -Filter 'nanoFramework.TestAdapter.dll' -ErrorAction SilentlyContinue |
    Sort-Object FullName -Descending | Select-Object -First 1
```

That reads the folder, not the checkout's references. So a `nanoFramework.TestFramework` version that is in `packages\` but referenced by nothing still wins if it sorts higher, and `Run-Tests.ps1` runs against an adapter this checkout does not reference. `CLAUDE.md` records what that class of mismatch costs: the `NFUnitTest` rename produced a green `vstest` run that executed nothing and went unnoticed for three commits.

#68 stops the folder being *polluted* by other checkouts, but does not clean what is already there and does not stop an ordinary version bump leaving the previous version behind. Right now only one `TestAdapter.dll` is present in the main checkout, so no wrong adapter is being picked — this is a mechanism confirmed by measurement, not an observed failure.

## What closing it would look like

Either end, or both:

- **Prune**: give `Restore-Packages.ps1` a way to report (and optionally remove) `packages\` folders no `packages.config` in this checkout references. It already computes the referenced set; the diff is the comparison and a switch. `Get-SmartHomePackagesConfig` (added in PR #77) is the list to compare against.
[...]

PR #86, opened 2026-09-01
feat(scripts): give the host-side script logic a test home

### LINKED ISSUE #74: Host-side script logic has no test home, so every fix to it needs a throwaway harness

### LINKED ISSUE #83: Conformance capture window hardcodes the lifecycle step count instead of deriving it
`Get-ConformanceCaptureSeconds` (`scripts/Run-IntegrationTests.ps1:1357`) sizes the managed-debug capture that runs alongside the Homie conformance check from the check's own timeouts. One of its terms is the number of lifecycle steps, written as a literal:

```powershell
$lifecycleSteps = 5
...
return $Settings.SettleSeconds +
       $Settings.RecoverySeconds +
       (($lifecycleSteps + 2) * $Settings.CommandTimeoutSeconds) +
       60
```

The actual steps are the five-element `$lifecycleSteps` array in `Measure-HomieConformance` (:1789):

```powershell
$lifecycleSteps = @(
    @{ Command = 'alert';    Expect = 'alert';    Refused = $false }
    @{ Command = 'sleeping'; Expect = 'alert';    Refused = $true  }
    @{ Command = 'ready';    Expect = 'ready';    Refused = $false }
    @{ Command = 'sleeping'; Expect = 'sleeping'; Refused = $false }
    @{ Command = 'ready';    Expect = 'ready';    Refused = $false }
)
```

Same fact, same name, spelled twice about 420 lines apart in one file, with nothing reconciling them.

## What goes wrong

Adding a lifecycle step — a sixth transition, or splitting the refused one so the two halves can be asserted apart — leaves the capture budget short by one `CommandTimeoutSeconds`, which is 30s on the current `HomieClientCheck` catalog entry.

That is quiet in the way this repository has been bitten by before. The capture is a backstop: `Stop-DeviceDebugCapture` closes it as soon as the check returns, so an undersized budget costs nothing on a fast run and takes the device-side log away exactly on the slow run that is worth reading. `Get-ConformanceCaptureSeconds`'s own comment says as much ("A capture that ends early takes the device-side evidence with it exactly when the run was slow enough to be worth reading"). Losing that log is the evidence gap #35 was about, and it would be re-opened by an edit that has nothing to do with capturing.

Nothing warns, because the capture ending is a normal outcome — the monitor is killed on every run either way.

## Status

Confirmed by reading `origin/main` (`8c725c7`). Not observed in a run: the literal and the array agree today, so no current run is undersized. This is a latent duplication, filed for the same reason #79 was — the mechanism is confirmed, the failure has not happened yet.

## What closing it would look like

Derive the count rather than restate it. The step table is a constant, so hoisting it to script scope and having `Get-ConformanceCaptureSeconds` read `.Count` is enough, and it removes the literal entirely.

Worth checking against #74 before starting: point 2 of that issue moves reusable helpers out of this file's body so they can be dot-sourced and tested, and the step table is exactly the kind of thing that moves with them. If #74 lands first this is a one-line change inside it; if not, it stands alone.

**Files:** `scripts/Run-IntegrationTests.ps1`

Found by a `/simplify` pass over `scripts\` on `main`, triaged against the queued work; deliberately not folded into any of it.

JojoEffect:
Closes #74. Also closes #83, which the backlog queue asked to fold in here if this landed first.

`scripts\` decides what the integration suite reports, and none of it had anywhere to be tested. Proving a change meant running the whole on-device suite or building a throwaway harness — four of those were written and deleted in a week (78 cases, none re-runnable), and each re-implemented the same trick: lift the functions out of `Run-IntegrationTests.ps1` with the PowerShell AST, because dot-sourcing it runs the suite.

## What is here

**The load-bearing part: the AST trick is no longer needed.** `Run-IntegrationTests.ps1` keeps its catalog and its functions at file scope and puts the run — reading `local.env`, the pre-flight, the broker, the deploy loop — behind a guard, so a dot-source defines everything and does nothing. The guard reads `$MyInvocation.InvocationName`, which is `.` for every spelling of a dot-source and the path or `&` for every spelling of a run, so the safe answer is the default: **a dot-source can never start a suite that flashes the device.** An `-AsLibrary` switch would have inverted that.

The catalog validation moved into `Get-CatalogValidationError`, which returns the first violation instead of writing it and exiting, so the rules can be asserted without a run.

**`scripts\Run-ScriptTests.ps1`** runs `scripts\tests\*.Tests.ps1`: **113 cases in ~6s**, needing nothing installed. A run that executed zero tests fails, on the same reasoning `Run-Tests.ps1` already carries.

**CI runs it first**, before the MSBuild and nanoFramework setup, because it needs none of it. It is the only automated coverage this repository has of the integration tooling itself — #22's unit tests run on the virtual device and cannot reach a PowerShell function.

## Two decisions worth reviewing

**Not Pester.** Pester 5 does not ship with Windows. What ships, and what is on this machine, is Pester **3.4.0**, whose dialect (`Should Be`, not `Should -Be`) is incompatible with anything written for 5. Adopting it would mean `Install-Module` on every desk, a pinned install step in CI, a row in `Test-Setup.ps1`, and a version guard at every import so PowerShell does not auto-load the in-box 3.4.0. The point of the suite is that proving a host-side change costs nothing, and a prerequisite that has to be installed first is a reason not to run it — which is why `Test-Setup.ps1` exists at all. The runner is ~200 lines of `Describe`/`It`/`Assert-*` this repository owns; `TestRunner.ps1` records the trade and what would make it worth revisiting.

**The guard, not hoisting helpers.** #74's point 2 offered either. The guard makes every one of the 32 functions reachable in one ~90-line move; hoisting would have grown `Common.ps1` past 1500 lines and mixed suite-specific logic into it. The duplication that hoisting would also have fixed is filed as #85 rather than folded in.

## What was verified

**Nothing ran on hardware. Nothing was deployed. The device was not touched** — it is still on `MqttReconnectCheck`, where the last suite run left it.

- `Run-ScriptTests.ps1`: **113 passed, 0 failed, exit 0**.
- **With both `local.env` files moved aside: 113 passed, exit 0.** That is the CI claim, checked rather than assumed.
- The failure paths, checked deliberately: a filter matching nothing exits 1 with "No tests ran"; a deliberately failing file exits 1 and reports the assertion and the exception with the right file and line.
- The dot-source guard, in a child process: 32 functions and the 5-entry catalog defined, `$repoRoot` undefined, `$LogDirectory` empty, no log directory created, in ~200ms. A second case walks the AST and asserts every top-level statement above the guard is a declaration, so an `Import-SmartHomeLocalEnv` or a `Test-Path` cannot drift back to the top of the file.
- `Run-IntegrationTests.ps1`'s two safe run paths still behave: `-Tests NoSuchTest` and `MqttReconnectCheck -NoBroker` report the same messages and exit 1, both before anything is built or flashed.
- `Test-Setup.ps1`: 16 checked, 0 failed.

**What a green run here does not claim.** `Start-HomieCapture`, `Wait-ForRetainedValue`, `Invoke-BrokerOutageCheck` and `Invoke-HomieConformanceCheck` are not covered — they own a process or a broker. Only `Run-IntegrationTests.ps1` on real hardware proves those, and this suite is not a substitute for saying so. That boundary is written into the skill and filed as #84.

## The review pass found real defects, in the runner itself

`/code-review high` over the first commit found three, all fixed in `d3ce4f4` with six regression cases:

- **`Assert-Equal` passed for a false claim.** PowerShell's comparison operators *filter* a collection rather than returning a boolean, so `@('a','b') -ceq 'a'` is `@('a')` — truthy — and `Assert-Equal -Expected @('a','b') -Actual 'a'` **passed**. It still failed for `-Actual 'z'`, which is what made it look like it worked: a test framework selectively unable to fail, which is the exact shape of the green-run-that-executed-nothing this repository has already shipped. Both operands are now refused with a message naming `Assert-ArrayEqual`.
- **`Assert-Match` had the same trap in the other direction** — `@('cat','dog') -notmatch 'cat'` is `@('dog')`, so it reported no match on output that plainly matched.
- **`Assert-Contains` was case-insensitive**, contradicting the case-sensitive default the rest of the runner keeps — and the claims already carried on it are about exact tokens (`127.0.0.1`, `%t %r %p`).

The suite also caught a bug in this branch while it was being written: `Get-CatalogValidationError`'s `$requiredKeys` local and its `$RequiredKeys` parameter are the same variable, because PowerShell names are case-insensitive, so the local assignment wrote a `string[]` into a `[hashtable]`-typed parameter. Renamed, with a comment.

## #83, folded in

`Get-ConformanceCaptureSeconds` sized the debug capture from a literal `$lifecycleSteps = 5`, about 420 lines from the five-element array `Measure-HomieConformance` walks. The table moved to file scope and both read it. `-LifecycleStepCount` is defaulted from it, which is what makes the fix provable: a test that recomputed the formula would pass against a literal too, so the case asserts that moving the count from 5 to 6 moves the budget by exactly one `CommandTimeoutSeconds`. Latent, not observed — every call today has five steps.

## Two known defects are pinned rather than endorsed

So their fixes have to come past a failing test: `Get-NanoFrameworkTestAdapterDir` picking `3.0.9` over `3.0.80` by text sort (#79 — the next queued item), and `Get-NfProjectAssemblyName` failing on a bracketed project path (#80).

The second turned out to fail differently from how #80 records it — at parameter binding, not as `ItemNotFound`, because `-Raw` is a FileSystem dynamic parameter resolved from the provider of the *resolved* path. [Corrected on #80](https://github.com/JojoEffect/SmartHome/issues/80#issuecomment-5498673479), since grepping for `ItemNotFound` would not find it.

## Filed, not fixed

- **#84** — the verdict functions that own a process or a broker are still uncovered, and `Test-Attribute` cannot be reached from a test at all (it is nested inside `Measure-HomieConformance` and closes over a `$snapshot` reassigned later in the same scope). That one decides whether the conformance assertions can be covered.
- **#85** — five duplications between `Run-IntegrationTests.ps1` and `Start-DevEnv.ps1` that a `/simplify` pass recorded on #74 on the assumption this change would absorb them. It did not, so they needed a home rather than closing with #74.

### DIFF
--- a/.claude/skills/smarthome-script-tests/SKILL.md
+++ b/.claude/skills/smarthome-script-tests/SKILL.md
@@ -56,11 +56,26 @@ dot-source defines the catalog and the functions and runs nothing. Do not go bac
 functions out of it with the PowerShell AST — a test that reads its subject through a parser can
 drift from what actually runs, which is what issue #74 was about.
 
-The vocabulary is in `scripts\tests\TestRunner.ps1`: `Describe`, `It`, `Assert-Equal`
-(case-sensitive; pass `-IgnoreCase` where case is not part of the claim), `Assert-ArrayEqual`,
-`Assert-True`/`Assert-False`, `Assert-Null`/`Assert-NotNull`, `Assert-Match`, `Assert-Contains`,
-`Assert-Throws`, plus `New-TestDirectory` and `Set-TestFileContent` for fixtures. Every assertion
-takes `-Because`, which is what the failure message leads with.
+The vocabulary is in `scripts\tests\TestRunner.ps1`: `Describe`, `It`, `Assert-Equal`,
+`Assert-ArrayEqual`, `Assert-True`/`Assert-False`, `Assert-Null`/`Assert-NotNull`,
+`Assert-Match`, `Assert-Contains`, `Assert-Throws`, plus `New-TestDirectory` and
+`Set-TestFileContent` for fixtures. Every assertion takes `-Because`, which is what the failure
+message leads with.
+
+Two things about them worth knowing before writing a case:
+
+- **Comparisons are case-sensitive** (`Assert-Equal`, `Assert-ArrayEqual`, `Assert-Contains`);
+  the first and last take `-IgnoreCase`. `Assert-Match` is a regex and keeps `-match`'s own
+  case-insensitive default.
+- **`Assert-Equal` and `Assert-Match` refuse a collection** rather than comparing one, and say
+  so. PowerShell's `-eq` and `-match` *filter* a collection instead of returning a boolean, so
+  `@('a','b') -ceq 'a'` is `@('a')` — truthy — and an assertion built on that would pass
+  whenever the expected collection merely *contained* the actual value. Use `Assert-ArrayEqual`
+  for collections, and join at the call site for `Assert-Match`.
+
+A test file may stub a function the subject calls (`Get-SmartHomeDevEnvPath`, say) by defining
+it after the dot-source; command lookup walks the calling scope chain, so the stub wins inside
+that file or that `Describe` and nowhere else.
 
 ## Why not Pester
 
--- a/.claude/skills/smarthome-script-tests/SKILL.md
+++ b/.claude/skills/smarthome-script-tests/SKILL.md
@@ -1,0 +1,82 @@
+---
+name: smarthome-script-tests
+description: Run the host-side script tests — the desk-provable half of scripts\, covering Common.ps1's helpers and the integration runner's verdict logic. Use before and after any change under scripts\, or when a host-side function needs proving without a device. For the on-device suite use smarthome-integration-tests instead.
+---
+
+# Run the host-side script tests
+
+```powershell
+.\scripts\Run-ScriptTests.ps1
+```
+
+No device, no broker, no network, no `local.env.ps1`, no restored `packages\`. About 6 seconds
+for ~100 cases. Exit 0 if everything passed, 1 otherwise — including when *nothing ran*, which is
+a failure on purpose: this repo already shipped three commits on a green `vstest` run that
+executed nothing.
+
+Useful switches:
+
+```powershell
+.\scripts\Run-ScriptTests.ps1 -File Common          # one subject
+.\scripts\Run-ScriptTests.ps1 -Name '*retained*'    # one claim, across files
+.\scripts\Run-ScriptTests.ps1 -Detailed             # every case, not just group counts
+```
+
+## What it covers, and what it deliberately does not
+
+`scripts\` decides what the integration suite reports, and it is the half of the suite a desk can
+exercise: `Get-CatalogValidationError`, `ConvertFrom-HomieCaptureLine`, `ConvertTo-HomieSnapshot`,
+`Get-HomieLivePayloads`, `Wait-ForAnnounceWitnessed`, `Get-SubscriberLogLineCount`, plus
+`Common.ps1`'s path globs, dev-environment state and the deploy-padding record.
+
+It does **not** cover anything that needs a broker, a subscriber process or the device —
+`Start-HomieCapture`'s connect wait, `Invoke-BrokerOutageCheck`, `Invoke-HomieConformanceCheck`.
+Those still need `Run-IntegrationTests.ps1` on real hardware. Running this suite is not a
+substitute for saying what was verified on hardware in a pull request.
+
+## Adding a test
+
+One file per subject, `scripts\tests\<Subject>.Tests.ps1`, dot-sourcing the subject at the top:
+
+```powershell
+Set-StrictMode -Version Latest
+$ErrorActionPreference = 'Stop'
+
+. (Join-Path (Split-Path -Parent $PSScriptRoot) 'Common.ps1')
+
+Describe 'Get-Something' {
+    It 'does the thing' {
+        Assert-Equal -Expected 'x' -Actual (Get-Something)
+    }
+}
+```
+
+`Run-IntegrationTests.ps1` can be dot-sourced the same way: a guard near the bottom of it means a
+dot-source defines the catalog and the functions and runs nothing. Do not go back to lifting
+functions out of it with the PowerShell AST — a test that reads its subject through a parser can
+drift from what actually runs, which is what issue #74 was about.
+
+The vocabulary is in `scripts\tests\TestRunner.ps1`: `Describe`, `It`, `Assert-Equal`
+(case-sensitive; pass `-IgnoreCase` where case is not part of the claim), `Assert-ArrayEqual`,
+`Assert-True`/`Assert-False`, `Assert-Null`/`Assert-NotNull`, `Assert-Match`, `Assert-Contains`,
+`Assert-Throws`, plus `New-TestDirectory` and `Set-TestFileContent` for fixtures. Every assertion
+takes `-Because`, which is what the failure message leads with.
+
+## Why not Pester
+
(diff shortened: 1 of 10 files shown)


JojoEffect:
CI green on `e24635d`: **113 passed, 0 failed in 27.4s** on the hosted runner, and the unit tests behind it 55 of 55 executed.

Worth stating explicitly, because a step that passes having run nothing is this repository's recurring failure: the run log shows all 24 groups and the final count, not just an exit code. It took 27.4s there against ~6.5s at a desk — the dot-source guard case spawns a child PowerShell, which is most of the difference on a hosted runner.

It ran first, before `setup-msbuild`, `setup-nuget`, the nanoFramework components and the `nanoclr` install, and needed none of them.

Issue #69, opened 2026-08-31
Restore-Packages.ps1 reports a MAX_PATH failure as a missing file, and leaves a partial package folder behind


JojoEffect:
`Restore-Packages.ps1` reports a MAX_PATH failure as "part of the path could not be found", which reads as a missing file

## What is wrong

`scripts\Restore-Packages.ps1` copies each package folder out of the local NuGet cache with

```powershell
Copy-Item -Path $cachePath -Destination $destPath -Recurse
```

When the destination exceeds Windows' 260-character path limit, `Copy-Item` fails with a `DirectoryNotFoundException` naming a file it was in the middle of writing. Nothing in the message mentions path length, and the file it names does not exist *because the copy is what would have created it* — so it reads as a corrupt or incomplete NuGet cache, which is a different problem with a different fix.

## Reproduction

Create a checkout whose path is long enough that `<root>\packages\nanoFramework.CoreLibrary.1.17.11\nanoframework.corelibrary.1.17.11.nupkg.sha512` exceeds 260 characters, and run the script. Observed (German MSBuild/PowerShell locale on this machine):

```
Copy-Item : Ein Teil des Pfades "...\packages\nanoFramework.CoreLibrary.1.17.11\nanoframework.corelibrary.1.17.11.nupkg.sha512" konnte nicht gefunden werden.
In ...\scripts\Restore-Packages.ps1:71 Zeichen:13
+             Copy-Item -Path $cachePath -Destination $destPath -Recurs ...
    + CategoryInfo          : WriteError: (nanoframework.c...11.nupkg.sha512:FileInfo) [Copy-Item], DirectoryNotFoundException
```

The failing path was 266 characters. The same checkout, moved to a shorter path, restores all 29 packages in 1.6s.

## Why it matters even though worktrees are shallow

The repo's own worktrees sit at `.claude\worktrees\<name>`, which is short enough. But the repository root is wherever the user cloned it, and the deepest path the restore has to write is roughly `<root>` + 100 characters (the package folder name is repeated inside it as the `.nupkg.sha512` filename). Roughly 155 characters of room for the checkout path is not a lot on Windows, and the failure it produces points at the wrong cause.

There is also a second, unrelated trigger for the same misreading: `Copy-Item -Recurse` failing partway through leaves a half-copied `packages\<Id>.<Version>\` folder behind, and the script's own already-present check is `Test-Path $destPath` — so the *next* run skips that package as restored when it is not.

## What closing it would look like

[...]

PR #87, opened 2026-09-02
fix(scripts): resolve the test adapter by reference and prune packages\

### LINKED ISSUE #78: Restore-Packages.ps1 aborts on a packages.config with no <package> children
`Restore-Packages.ps1` reads each config's packages with a property access:

```powershell
[xml]$xml = Get-Content $config.FullName
foreach ($package in $xml.packages.package) {
```

Under `Set-StrictMode -Version Latest`, which the script sets at the top, `$xml.packages.package` **throws** when the `<packages>` element has no `<package>` children — the property does not exist on that object. `$ErrorActionPreference` is `Stop`, so the restore aborts on that file rather than skipping it.

`Test-Setup.ps1` already hit this and guards against it, with a comment saying exactly why:

> XPath rather than `$xml.packages.package`: under `Set-StrictMode -Version Latest` the property access throws on a `packages.config` with no `<package>` children, which would abort the whole preflight. `SelectNodes` just returns an empty list.

`Restore-Packages.ps1` is the other reader of these files and did not get that treatment — the same asymmetry as #68, in the same pair of scripts.

## Reproduction

Point the script at a checkout containing a `packages.config` whose root element is empty:

```
<packages></packages>
```

Observed (Windows PowerShell 5.1, German diagnostics on this machine):

```
Restoring packages referenced by 2 packages.config file(s)...
Restore-Packages.ps1 : Die Eigenschaft "package" wurde für dieses Objekt nicht gefunden.
    + FullyQualifiedErrorId : PropertyNotFoundStrict,Restore-Packages.ps1
```

The script exits without restoring anything, and without naming the file it choked on.

## Impact

Latent today: every `packages.config` in this repo currently declares at least one package, so nothing triggers it. It becomes reachable the moment a project is added with its dependencies stripped, or a config is hand-edited down to nothing while pinning versions — which is exactly the workflow the script's own documentation names as a reason to run it. The failure is loud but misleading: it reads as a script defect rather than "this config is empty".

Found while fixing #68 (PR #77), where a synthetic fixture with empty configs reproduced it. Not fixed there, to keep that diff to the glob it was about.

## What closing it would look like

Match `Test-Setup.ps1`: read the packages with `$xml.SelectNodes('/packages/package')`, which returns an empty list rather than throwing. Both scripts then parse these files the same way, as they now glob for them the same way.

### LINKED ISSUE #79: packages\ is never pruned, and the test adapter is picked by sort order rather than by reference

JojoEffect:
Closes #79. Closes #78 — folded in rather than filed alongside, because both halves of #79 need the same parse and fixing it twice was the alternative (the reasoning is in [#79's own comment](https://github.com/JojoEffect/SmartHome/issues/79#issuecomment-5498612053)).

## What was wrong

`Get-NanoFrameworkTestAdapterDir` resolved the unit-test adapter by taking the **highest-sorting** `nanoFramework.TestAdapter.dll` anywhere under `packages\`. That sort is textual, so it is not a version order, and nothing has ever pruned `packages\` — `Restore-Packages.ps1` only ever added to it. So an ordinary version bump leaves the previous version's folder behind and the suite can run against an adapter this checkout does not reference.

Measured on the main checkout while working this: **61 folders present, 29 versions referenced, 32 stale.**

Reproduced rather than argued. With `nanoFramework.TestFramework.3.0.9` planted beside the referenced `3.0.80`, on the same `packages\`, side by side:

```
old resolver would pick: ...\packages\nanoFramework.TestFramework.3.0.9\build
new resolver picks:      ...\packages\nanoFramework.TestFramework.3.0.80\lib\net48
```

`CLAUDE.md` records what that class of mismatch costs — the `NFUnitTest` rename produced a green `vstest` run that executed nothing, unnoticed for three commits — so the failure mode to design against is "picked something", not "picked nothing".

## What changed

**One parse, shared (`scripts\Common.ps1`).** `Get-SmartHomeReferencedPackage` returns one record per distinct reference (`Id`, `Version`, `Name`, expected `Path`). `Restore-Packages.ps1` and `Test-Setup.ps1` computed that set separately, differing only in the XML access — and that difference *was* #78: `$xml.packages.package` throws under `Set-StrictMode -Version Latest` on a `packages.config` whose `<packages>` has no `<package>` children, aborting the restore without naming the file. The preflight had already been given XPath for exactly that reason. Both ends of the same reproduction, run against this branch:

```
old parse threw: Die Eigenschaft "package" wurde für dieses Objekt nicht gefunden.
new: "Restoring 1 package(s) referenced by this checkout..."  (the real reference beside the empty config is read)
```

The same reasoning one level down: attributes are read with `GetAttribute` rather than as properties, so a `<package>` missing an `id` or a `version` is warned about and skipped instead of throwing — never silently, because a dropped reference is a package that then looks restored to everything downstream.

**Resolve by reference, not by sort.** `Get-NanoFrameworkTestAdapterDir` now demands the `nanoFramework.TestFramework` version the checkout declares. It returns `$null` — which both callers already treat as "cannot run the unit tests" — with a warning naming which of three reasons it is: the referenced version is not restored, nothing references the framework at all, or two projects reference different versions. It deliberately does **not** fall back to a restored version nothing references; that fallback is the thing that made it lie. `Test-Setup.ps1` silences the warning and builds its row from the same list, so the preflight table stays a table.

**Prune.** `Restore-Packages.ps1` ends every run with a one-line count of the `packages\` folders nothing references. `-Prune` removes them; `-Prune -WhatIf` lists them without touching anything. Off by default because the script is called unattended by `Initialize-Worktree.ps1`, which asked for a checkout to be filled in, not for anything to be deleted. It **refuses** to prune when the checkout references nothing at all — the complement of an empty reference set is the whole folder, and that is a broken checkout rather than a clean one.

The set difference lives in `Get-SmartHomeUnreferencedPackageDir`, which takes the referenced records as a parameter rather than a repo root: the caller already has them, and a destructive step that depends only on its arguments can be proved at a desk. That is the same seam #86 argued for, applied to the one block in this change that deletes things.

`Restore-Packages.ps1`'s "already present" count is now a count of packages rather than of mentions — 29, not 130, for this repository's 14 configs.

## Verification

**Desk — `.\scripts\Run-ScriptTests.ps1`: 130 passed, 0 failed in ~7s** (113 before this change, so +17 cases). No device, no broker, nothing installed.

The case that pinned the lexical sort is **inverted rather than deleted** — same fixture, `3.0.9` and `3.0.80` both restored, and now the referenced version wins. Alongside it: refusing to guess between two referenced versions, refusing to fall back to a restored version nothing references, and the `#78` empty-`<packages>` case.

**Hardware — `.\scripts\Run-Tests.ps1` on the ESP32: 55 of 55 executed, 55 passed, exit 0.** That is the claim the desk tests cannot make: the adapter resolved by reference is one `vstest.console` actually loads and runs tests through.

```
Adapter: ...\packages\nanoFramework.TestFramework.3.0.80\lib\net48
Results: 55 passed, 0 failed, 55 of 55 executed.
Tests passed on real hardware.
```

**Virtual — `.\scripts\Run-Tests.ps1 -RunSettings nano.ci.runsettings`: 55 of 55 executed, exit 0**, same adapter path. This is the path CI takes.

Also exercised by hand on real checkouts, not only on fixtures: the restore into a fresh worktree (29 packages), the report line, `-Prune -WhatIf`, `-Prune` removing exactly the two planted stale folders and keeping the referenced ones, the preflight's OK row, and its FAIL row with the referenced version renamed away (`nanoFramework.TestFramework 3.0.80 referenced, its adapter not in packages\`). The refuse-to-prune guard was exercised in a throwaway checkout that references nothing: exit 1, both folders still there.

**The device now carries the unit-test assembly**, not `MqttReconnectCheck` and not RoomSensor. Redeploy RoomSensor when it should go back to its real job.

## Not in this change

- One stale comment corrected in passing, in a file this diff rewrites anyway: `Restore-Packages.ps1`'s exit-code note named `Run-IntegrationTests.ps1` and `Start-DevEnv.ps1` as its callers. Neither invokes it; `Initialize-Worktree.ps1` is the only script that does (`grep` over `scripts\`). Not filed as an issue — it is two wrong script names in a comment three lines from new ones.
- No prune row was added to `Test-Setup.ps1`. It would `WARN` on 32 folders for everyone until they prune, and the preflight is meant to be actionable.


JojoEffect:
CI green on the hosted runner, with the counts rather than a bare exit code — a step that passes having run nothing is this repository's recurring failure:

```
Run host-side script tests        130 passed, 0 failed in 29.1s
Run unit tests on the virtual device
  Adapter: D:\a\SmartHome\SmartHome\packages\nanoFramework.TestFramework.3.0.80\lib\net48
  Results: 55 passed, 0 failed, 55 of 55 executed.
```

Worth being precise about what that second line does and does not prove. CI restores with `nuget restore SmartHome.sln` onto a fresh runner, so `packages\` holds only the referenced versions and the old sort-order resolver would have picked the same folder there. The fix itself is proved by the desk cases and by the local run with `3.0.9` planted beside `3.0.80`; this run confirms the new path resolves and executes on a machine that is not mine.

JojoEffect:
`/code-review high --fix` over `906c799`. Five findings, **all five applied** in `d4ac528`.

Correction to that commit's own message, which is wrong about this and cannot be fixed without rewriting a pushed branch: it says "five findings, four of them applied" and names a fifth about `Remove-Item` failing partway through a `-Prune`. That was considered during the review and deliberately *not* reported as a finding — a partial prune of folders that are all unreferenced is harmless and the abort is loud — so the correct count is five reported, five applied. The `Remove-Item` behaviour is unchanged either way, so nothing in the diff turns on it.

The two that were real defects rather than wording:

**1. `-WhatIf` reported a restore it did not make.** `SupportsShouldProcess` applies to the whole script, not only to `-Prune`, so `Copy-Item` was suppressed while the loop still counted it. Measured on a checkout whose `packages\` was missing both its references:

```
  restored: nanoFramework.CoreLibrary.1.17.11
  restored: nanoFramework.M2Mqtt.5.1.221
0 already present, 2 restored from local NuGet cache.
All referenced packages are available.
exit=0
packages\ folders actually on disk afterwards: nanoFramework.M2Mqtt.5.1.146
```

Nothing had been restored. `-Prune -WhatIf`, the run the docstring recommends, was the run that produced it. Now:

```
0 already present; -WhatIf, so nothing was restored.
-WhatIf: every referenced package was located, and nothing was written.
```

**2. The adapter row could abort the whole preflight.** `Get-NanoFrameworkTestAdapterDir` had started globbing every `packages.config` in the checkout, and `Test-Setup.ps1` called it without the `-ErrorAction SilentlyContinue` it passes for the `packages\` row two blocks above. A subtree the enumerator cannot read — a denied folder, a junction loop, a `packages\` path over MAX_PATH (#69 records that failure in this exact folder) — would raise a non-terminating error that `$ErrorActionPreference = 'Stop'` turns terminating, killing the preflight before it printed its table. That is precisely the "report rather than abort" contract the script's own docstring states.

Same finding measured as an efficiency one: the preflight had gone from one recursive glob of the checkout to three, and one pass over the main checkout costs ~2.4s (it covers `packages\` and every linked worktree's source tree).

Both fixed by the same seam `Get-SmartHomeUnreferencedPackageDir` already uses — take the list as a parameter. `Get-SmartHomeReferencedPackage` gained `-Config`, `Get-NanoFrameworkTestAdapterDir` gained `-ReferencedPackage`; omitting either globs exactly as before, and `Test-Setup.ps1` is back to one glob.

The remaining three were accuracy, not behaviour: a comment justifying case-insensitive folder matching with the NuGet *cache*'s lowercase layout rather than anything ever seen in `packages\`, and a `-Because` that counted configs where it meant references.

**134 desk cases, 0 failed** (130 before), including both the handed-over and the glob-for-itself path of each new parameter. `Restore-Packages.ps1`, `Test-Setup.ps1` (16 rows, all OK) and the resolver re-run by hand on the real checkout after the change; the adapter still resolves to `packages\nanoFramework.TestFramework.3.0.80\lib\net48`, byte-identical to what the hardware run exercised, so that run was not repeated.

Issue #19, opened 2026-08-22
dotnet run re-evaluates the monitor project on every capture


JojoEffect:
`Watch-DeviceDebugOutput.ps1` invokes `dotnet run --project ... --no-build` for every
capture. `--no-build` skips compilation, but `dotnet run` still performs a full MSBuild
*evaluation* to resolve the output path — roughly 0.8–1.5s per invocation, several times
per suite run including retries.

The `-BuildOnly` pass the runner already makes has produced that output. Resolve the
path once (`dotnet build ... -getProperty:TargetPath`, or glob
`tools\DeviceDebugMonitor\bin\<Config>\net*\DeviceDebugMonitor.exe`) and invoke the exe
directly for each capture.

**Files:** `scripts/Watch-DeviceDebugOutput.ps1`, `scripts/Run-IntegrationTests.ps1`

JojoEffect:
A second call site for this, added by PR #91 (issue #46).

`Deploy-ToDevice.ps1` now clears the device's deployment area before every flash, through
`Clear-SmartHomeDeviceDeployment` in `Common.ps1`, which does the same pair this issue is
about: a `dotnet build --nologo -v quiet` and then a `dotnet run --project ... --no-build`.
So the per-invocation cost is now paid once per *deploy* as well as once per capture — six
more times in a `Run-IntegrationTests.ps1` suite run, on top of the captures.

Measured while verifying #46 on hardware: the whole erase call is 2.8s when the device has
nothing to erase, of which the device conversation is a small part — the rest is the build
and the project evaluation this issue names.

A `-NoBuild` switch was deliberately *not* added to that function to work around it. The
runner's existing `-BuildOnly` pass has already produced the output by then, so the fix
described here — resolve the built exe once and invoke it directly — covers the deploy path
and the capture path with one change and no switch for a caller to remember.

**Files**, updated: `scripts/Watch-DeviceDebugOutput.ps1`, `scripts/Run-IntegrationTests.ps1`,
`scripts/Common.ps1` (`Clear-SmartHomeDeviceDeployment`, `Get-SmartHomeDeviceMonitorProject`).

PR #91, opened 2026-09-02
fix(deploy): ask the device what to erase instead of guessing

### LINKED ISSUE #46: Deploy padding cannot see flashes made outside Deploy-ToDevice.ps1
Follow-up to #17 / PR #45, which replaced `Deploy-ToDevice.ps1`'s flat 400KB pad with a
per-COM-port record of the last flashed size.

The record only describes what **that script** flashed. Two other things write the same
deployment partition and, like `nanoff`, erase only their own footprint —
`DeploymentExecuteFull` in `nf-debugger`'s `WireProtocol/Engine.cs` erases exactly the
assemblies' combined length:

- The **nanoFramework test adapter**, reached through `Run-Tests.ps1`. Handled: that script
  now clears the record when its run settings target real hardware.
- **Visual Studio's F5 deploy**. Not handled, and not hookable from this repo.

So after someone deploys from Visual Studio, the record still names whatever
`Deploy-ToDevice.ps1` last flashed. If the VS-deployed image was larger, the next scripted
deploy pads to the smaller recorded number and leaves the VS image's tail unerased — and the
CLR does not stop at it, because `ContiguousBlockAssemblies` (`nf-interpreter`,
`src/CLR/Startup/CLRStartup.cpp`) walks the whole deployment partition and `continue`s past a
header that fails its check rather than breaking. That is exactly the stale-assembly failure
the padding exists to prevent.

The current mitigation is documentation plus a manual `-FullPad` switch, in `CLAUDE.md`, the
`smarthome-deploy` skill and the `device-ops` agent. Nothing enforces it, so the failure mode
is "someone forgets", and the symptom (a device loading two apps' assemblies) is confusing to
diagnose from scratch.

Worth noting the flat 400KB pad had the same blind spot in principle; it only masked this in
practice because it happened to exceed every image in the repo. This issue is about closing
the gap properly, not about a regression.

**Directions worth exploring, roughly in order of appeal:**

1. Read back what is actually on the device before deciding the pad. `tools/DeviceDebugMonitor`
   already speaks WireProtocol via `nanoFramework.Tools.Debugger.Net` and already has a
   `--dump-config` mode, so a `--deployment-length` mode that reports the extent of the
   deployed assemblies looks feasible. That would make the pad self-verifying and retire the
   record and `-FullPad` entirely.
2. Fingerprint the deployment region cheaply (a hash of the first N bytes, stored in the
   record) and fall back to the flat pad when it does not match.
3. Accept the gap and make it louder — e.g. have the deploy print the recorded image name it
   is padding against, so a mismatch with what the user knows they last flashed is visible.

**Files:** `scripts/Deploy-ToDevice.ps1`, `scripts/Common.ps1` (deploy-state section),
`tools/DeviceDebugMonitor/Program.cs`

JojoEffect:
Closes #46.

`Deploy-ToDevice.ps1` padded every image with 0xFF as far as a per-COM-port record of what
*it* had last flashed, so a smaller app could not leave a larger one's tail where the CLR
still finds it. The record could only ever describe this script's own flashes: Visual
Studio's F5 deploy and the nanoFramework test adapter write the same partition over the
debugger connection and erase only their own footprint too. `Run-Tests.ps1` invalidated the
record for the adapter, Visual Studio could not, and `-FullPad` existed for the case only a
human could notice.

The device is asked now, immediately before the flash. Nothing on the host has to know what
is already there, so nothing can be wrong about it.

## What the device will and will not answer

Read out of the pinned sibling checkouts first, then measured on the device — and the two
disagreed on one point, which is why there are two commits.

| Route | Result |
|---|---|
| `Monitor_DeploymentMap` — the command whose name promises the used extent | **Stub in the firmware.** Replies with an empty payload (`Debugger.cpp`), so `GetDeploymentMap()` always returns an empty list |
| `AccessMemory_Read` over the deploy partition | **Refused.** `CheckPermission` lists no `BLOCKTYPE_DEPLOYMENT` case for reads. (`-DumpConfig` reads the *config* partition, which is permitted — that is why that one works) |
| `Debugging_Deployment_Status` | **Reads as the direct question, does not work here.** Answers with no usable geometry on this ESP32. Nothing in `nf-debugger`'s own ESP32 path depends on it either: `DeploymentExecuteFull` is its only caller, the ESP32 reports `IncrementalDeployment`, so `DeploymentExecuteIncremental` runs instead |
| `Monitor_FlashSectorMap` | **Works.** Its `c_MEMORY_USAGE_DEPLOYMENT` entry gives the partition's start and size |
| `AccessMemory_Erase` | **Permitted**, and the firmware skips the erase when the region from the given address to the end of the partition is already blank (`Esp32FlashDriver_IsBlockErased`). Same command Visual Studio's own deploy issues before it writes |

So nothing reports the used extent. The device is told to erase rather than asked what is
there — which is the same invariant, decided by the device.

## The change

- `DeviceDebugMonitor` gains `--erase-deployment <keep-bytes>`: reports the partition's
  geometry and has the device blank everything past the image about to be written. It stops
  the CLR first (`GetExecutionMode` / `PauseExecution`), the same sequence VS's deploy uses,
  because the CLR executes assemblies straight out of that memory-mapped flash.
- `Common.ps1`'s "Deploy state" section becomes `Clear-SmartHomeDeviceDeployment` plus
  `Read-SmartHomeDeploymentGeometry` — the parse of the monitor's one machine-readable line,
  split out so it can be proved at a desk. Everything else there needs a device.
- Gone: the record (`Get`/`Save`/`Clear-SmartHomeDeployState`), `-FullPad`,
  `-FallbackPadSize`, `Run-Tests.ps1`'s invalidation, and its second parse of
  `nano.runsettings` with it.
- The 0xFF pad survives for exactly one case: the device could not be reached at all (VS
  holding the debugger connection is the usual reason). The image is then padded to the whole
  partition and the flash goes ahead **with a warning** — slower, same guarantee. That case is
  not hypothetical; it is what the first hardware run did, before the geometry route was
  fixed, and it deployed correctly.
- The geometry the same call returns is ground truth for two constants this script had only
  ever carried as boot-log measurements. A **deploy address that disagrees with the device is
  now refused** rather than flashed — that mismatch is what once put every deploy into the
  *factory* partition, where `nanoff` reported success and the app silently never ran. A
  disagreeing partition *length* is a warning and the device's value wins.

The fit check runs **before** the erase, on purpose: the erase takes the device's current app
with it, so an image that was never going to fit is refused while the device still has
something to run.

## Verified on hardware (COM3, ESP32-D0WD-V3)

The device confirmed both constants: `DEPLOYMENT start=0x001E0000 length=1835008`, against
`-DeployAddress 0x1E0000` and `-DeployPartitionSize 0x1C0000`.

Run in this order, so the middle step reproduces #46's actual hole — a flash made by
something other than this script:

1. **`Deploy-ToDevice.ps1` (RoomSensor)** — took the erase path, flashed the bare 260132-byte
   image in 1.8s. The run before it, on the broken geometry route, had padded to 1835008 and
   taken 4.9s; both produced a working device.
2. **`Run-Tests.ps1` on the ESP32 — 57 of 57.** This leaves `NFUnitTest` on the device,
   flashed by the test adapter. Nothing cleared any record, because there is none.
3. **The erase, timed twice against that.** First call 3.7s, second 2.8s — the first erased,
   the second found it already blank. Then `Watch-DeviceDebugOutput.ps1` on the erased device:
   zero assemblies, zero of every table, `Error: a2000000`. The adapter's app was gone, which
   is the erase doing the thing the record could not have known to ask for.
4. **`Run-IntegrationTests.ps1 -Tests WifiCheck,MqttCheck,Bmp280Check` — all 3 PASS**, 72s.
   Every deploy took the erase path, no padding fallback, no warnings. `MqttCheck` is 142056
   bytes and `Bmp280Check` right after it is 132516 — smaller after larger, the stale-tail case
   itself.
5. **`Run-Tests.ps1 -RunSettings nano.ci.runsettings` — 57 of 57** on the nanoclr virtual
   device. CI's path.
6. **`Run-ScriptTests.ps1` — 146 passed, 0 failed**, ~7s, no device. Was 147: the 11
   deploy-record cases are gone and 10 cover the geometry parse and the monitor project
   resolution.

The suite costs ~3s per deploy more than before (72s against the 63s the same three tests took
on `origin/main`) — one debugger connection each. In exchange no deploy depends on anything
remembered.

The two host-decided integration tests were not run: they emit no marker and this change does
not touch their path. The device is left on `Bmp280Check`.

### DIFF
--- a/.claude/agents/device-ops.md
+++ b/.claude/agents/device-ops.md
@@ -20,16 +20,17 @@ Scripts you own (each has a matching `.claude/skills/smarthome-*` skill too):
 - Don't rewrite `Start-DevEnv.ps1`'s child launches to use `-RedirectStandardOutput`: that forces
   handle inheritance and makes piping the script's output hang forever. The ShellExecute +
   `log_dest file` + `cmd.exe` wrapper arrangement is deliberate and commented in the script.
-- `scripts\Deploy-ToDevice.ps1 [-Project <path>] [-Configuration Debug|Release] [-DeployAddress <hex>] [-FullPad]`
-  — build + flash. `-FullPad` pads the flat 400KB worst case instead of sizing the 0xFF pad
-  from the per-COM-port record of the last flash — use it on the first deploy after someone
-  has deployed from Visual Studio, which writes the deployment partition behind this script's
-  back. `-DeployAddress` defaults to this device's current real `deploy` partition
+- `scripts\Deploy-ToDevice.ps1 [-Project <path>] [-Configuration Debug|Release] [-DeployAddress <hex>]`
+  — build + flash. Before the flash it has the device erase anything past the new image's end,
+  over the debugger connection, so a Visual Studio deploy or a hardware unit-test run in between
+  needs no switch and nothing remembered. If it warns that it could not clear the deployment
+  area, Visual Studio is usually holding that connection; the flash still goes ahead, padded to
+  the whole partition. `-DeployAddress` defaults to this device's current real `deploy` partition
   offset (`0x1E0000`) — `nanoff`'s own hardcoded default (`0x1B0000`) landed inside the
-  **factory** partition instead and silently produced a deploy the CLR could never load. If a
-  deploy "succeeds" but the device never does anything afterward, verify with
-  `Watch-DeviceDebugOutput.ps1` before assuming it's an application bug — check for
-  `CLR_E_WRONG_TYPE` / zero-assembly resolution, which means this address is stale again.
+  **factory** partition instead and silently produced a deploy the CLR could never load. The
+  script now cross-checks that address against what the device reports and refuses to flash on a
+  mismatch, so a stale address surfaces as a refusal rather than as a device that boots and does
+  nothing.
 - `scripts\Run-Tests.ps1` — build + run the `SmartHome.UnitTests` unit suite on hardware.
 - `scripts\Run-IntegrationTests.ps1 [-Tests <names>] [-NoBroker]` — the whole `src\integrationTests`
   suite in one call: starts a detached broker, then per test deploys, captures managed debug
--- a/.claude/skills/smarthome-deploy/SKILL.md
+++ b/.claude/skills/smarthome-deploy/SKILL.md
@@ -13,15 +13,24 @@ hard way), then flashes via `nanoff`.
 .\scripts\Deploy-ToDevice.ps1                                                    # RoomSensor, Debug
 .\scripts\Deploy-ToDevice.ps1 -Project src\devices\IrrigationControl\IrrigationControl.nfproj
 .\scripts\Deploy-ToDevice.ps1 -Configuration Release
-.\scripts\Deploy-ToDevice.ps1 -FullPad                                           # after a VS deploy
 ```
 
-The image is padded with 0xFF far enough to erase whatever the previous deploy left behind —
-sized from a per-COM-port record of the last flash, falling back to a flat 400KB whenever that
-record can't be trusted. **Pass `-FullPad` on the first deploy after anyone has deployed from
-Visual Studio**, since VS writes the deployment partition without going through this script and
-the record then describes an image that is no longer on the device. `Run-Tests.ps1` clears the
-record itself, so no `-FullPad` is needed after a unit-test run.
+`nanoff` erases only as much as it writes, so a smaller app flashed after a larger one would
+leave the larger one's tail where the CLR still finds and loads it. The script has the **device**
+erase anything past the new image's end, over the debugger connection, immediately before the
+flash — so a deploy from Visual Studio or a hardware unit-test run in between needs nothing
+remembered and no switch. There is no per-machine record any more, and no `-FullPad`.
+
+Two things it prints that are worth reading rather than scrolling past:
+
+- **"Could not clear the deployment area"** — the device would not talk, usually because Visual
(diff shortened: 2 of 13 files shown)


JojoEffect:
Two things added after the hardware run, and one consequence worth calling out explicitly.

**`dff2274` drops a `-NoBuild` switch nothing passed.** `Clear-SmartHomeDeviceDeployment`
had one for a caller that does not exist. An unused switch is worse than the ~2s no-op build
it would have saved, and the saving is already owned by #19 — `dotnet run --no-build` still
pays a full project evaluation, and this change adds a second call site to that. Commented
there rather than filing a duplicate; no new issue.

**A behaviour change reviewers should see: the device now has no app between the erase and
the flash.** The old padding never removed anything until `nanoff` wrote over it, so a
failed flash left the previous app running. On ESP32 the erase takes the whole partition
(`Esp32FlashDriver_EraseBlock` ignores the address), so a `nanoff` failure now leaves the
device empty rather than stale. That is the direction to fail in — empty is loud and one
redeploy away, stale is the silent two-app failure this whole mechanism exists to prevent —
but it is a change, and it is why the erase runs immediately before the flash and why the
fit check runs before the erase. Both are commented in `Deploy-ToDevice.ps1`.

`Watch-DeviceDebugOutput.ps1` on the erased device is what that looks like, captured during
verification: zero assemblies, zero of every table, `Error: a2000000`.

PR #92, opened 2026-09-02
refactor(tests): lift the conformance attribute check out of its closure

JojoEffect:
First slice of #84 — deliberately that slice only. The issue names it as the thing to do first
because it is what decides whether the conformance assertions can ever be covered.

## What was wrong

`Test-Attribute` was defined *inside* `Measure-HomieConformance` and closed over that scope's
`$snapshot`, which the `/set` round-trip reassigns further down the same function. Every call
happened before that reassignment, so no assertion was ever measured against the wrong store —
but that was a property of where in a 500-line function the calls were written rather than of
anything the code said. And it could not be called from a test at all, which is why the largest
verdict function in the repository, and the one with two prior defects that passed while lying
(#34, #36), had no host-side coverage of any kind.

## What changed

`Get-AttributeFailure`, at file scope, taking the snapshot as a parameter and *returning* what is
wrong with the attribute instead of appending to a list it can only see from an enclosing scope.

The 19 call sites collect with `$script:conformanceFailures += …`, which is the idiom every other
assertion in that function already uses. The attribute assertions were the only ones hiding their
append inside a closure, so this is a consistency gain rather than added noise.

Renamed on the way out: a `Test-` verb returning strings reads as a boolean predicate.
`Get-CatalogValidationError` is the file's precedent for "returns what is wrong". Three comments
elsewhere in the file that named `Test-Attribute` were updated with it.

Behaviour is unchanged. The early return on a missing topic, the retain-flag check, the payload
comparison and `-AnyValue`'s meaning are all as they were.

## Tests

11 new cases in `scripts\tests\RunIntegrationTests.Tests.ps1`, built on the real
`ConvertTo-HomieSnapshot` rather than hand-rolled hashtables so a change to the entry shape breaks
them instead of leaving them asserting against a shape the capture no longer produces:

- a clean attribute reports nothing
- a missing topic is reported, and nothing further is asserted about it
- an attribute the broker did not replay as retained
- a payload mismatch, naming both values
- both faults of one attribute reported together rather than the first
- `-AnyValue` skips the payload comparison, still asserts the retain flag, still reports a missing topic
- the snapshot read is the one passed in, not one the caller happens to have in scope
- a clean attribute appends nothing to a caller's `+=`

The snapshot case was checked to be load-bearing by deleting the `-Snapshot` parameter: it is the
case that then fails. Worth recording that a smaller mutation — lowercasing `$Snapshot` to
`$snapshot` in the body — does **not** fail, because PowerShell variable names are case-insensitive
and it still binds to the parameter. The test's comment says so rather than claiming a mechanism
it does not have.

`.claude\skills\smarthome-script-tests\SKILL.md` records the new coverage and that this is the
first piece of the conformance verdict covered at all.

## Verification

**Desk:** `Run-ScriptTests.ps1` — **157 passed, 0 failed** (was 146 on `origin/main`).

**Hardware:** the full integration suite on the ESP32 on COM3 with a local Mosquitto —
**all 5 PASS, 200s**.

| Test | Result | Time |
|---|---|---|
| WifiCheck | PASS | 26s |
| MqttCheck | PASS | 23s |
| Bmp280Check | PASS | 25s |
| **HomieClientCheck** | **PASS** | 71s |
| MqttReconnectCheck | PASS | 55s |

`HomieClientCheck` is the one that matters here: 53 topics in the retained store, and every
rewritten call site produced a clean verdict — "attributes, datatypes, retained flags, /set
round-trip, refused out-of-format payloads, alert/sleeping, a refused transition and re-announce
all conform". Phase breakdown was normal (announce 10.1s, `/set` 3.8s, out-of-format 4.3s,
lifecycle 17.8s, re-announce 13.8s).

The device is left on `SmartHome.IntegrationTests.MqttReconnectCheck`, as a suite run always leaves
it. Redeploy RoomSensor when it should go back to its real job.

## Filed, not fixed here

Extracting the function put its comparisons under a lens: both the topic lookup and the payload
comparison are case-insensitive, so a device publishing `$state` as `READY` passes an assertion
that wants `ready`, and two topics differing only in case collapse to one snapshot entry. Measured
against the shipped function, not reasoned about. Pre-existing and out of this slice's scope, so
it is an issue rather than a rider on this diff.

## Not in this change

The rest of #84 — `Start-HomieCapture`/`Stop-HomieCapture`, `Wait-ForRetainedValue`,
`Test-DeviceConstant`, the log-reading half of `Invoke-BrokerOutageCheck` — needs the scripted
fake-subscriber technique and is its own project. #84 stays open.

One item of #84 is now moot rather than done: the `Clear-SmartHomeDeployState` temp-directory seam
it asks for went away with the whole deploy record in #91. Noted on the issue.

Part of #84.

### DIFF
--- a/.claude/skills/smarthome-script-tests/SKILL.md
+++ b/.claude/skills/smarthome-script-tests/SKILL.md
@@ -26,14 +26,19 @@ Useful switches:
 
 `scripts\` decides what the integration suite reports, and it is the half of the suite a desk can
 exercise: `Get-CatalogValidationError`, `ConvertFrom-HomieCaptureLine`, `ConvertTo-HomieSnapshot`,
-`Get-HomieLivePayloads`, `Wait-ForAnnounceWitnessed`, `Get-SubscriberLogLineCount`, plus
-`Common.ps1`'s path globs, dev-environment state and the deployment-geometry parse the deploy
-cross-checks its flash address against.
+`Get-HomieLivePayloads`, `Get-AttributeFailure`, `Wait-ForAnnounceWitnessed`,
+`Get-SubscriberLogLineCount`, plus `Common.ps1`'s path globs, dev-environment state and the
+deployment-geometry parse the deploy cross-checks its flash address against.
 
 It does **not** cover anything that needs a broker, a subscriber process or the device —
 `Start-HomieCapture`'s connect wait, `Invoke-BrokerOutageCheck`, `Invoke-HomieConformanceCheck`.
 Those still need `Run-IntegrationTests.ps1` on real hardware. Running this suite is not a
 substitute for saying what was verified on hardware in a pull request.
+
+`Get-AttributeFailure` is the first piece of the conformance verdict that is covered here at all.
+It was nested inside `Measure-HomieConformance` and closed over that scope's snapshot until issue
+#84 gave it the snapshot as a parameter; the rest of that function still needs a device, and the
+uncovered remainder is what #84 tracks.
 
 ## Adding a test
 
--- a/.claude/skills/smarthome-script-tests/SKILL.md
+++ b/.claude/skills/smarthome-script-tests/SKILL.md
@@ -26,18 +26,20 @@ Useful switches:
 
 `scripts\` decides what the integration suite reports, and it is the half of the suite a desk can
 exercise: `Get-CatalogValidationError`, `ConvertFrom-HomieCaptureLine`, `ConvertTo-HomieSnapshot`,
-`Get-HomieLivePayloads`, `Get-AttributeFailure`, `Wait-ForAnnounceWitnessed`,
-`Get-SubscriberLogLineCount`, plus `Common.ps1`'s path globs, dev-environment state and the
-deployment-geometry parse the deploy cross-checks its flash address against.
+`Get-HomieLivePayloads`, `Get-AttributeFailure`, `Get-ConformanceCaptureSeconds`,
+`Wait-ForAnnounceWitnessed`, `Get-SubscriberLogLineCount`, plus `Common.ps1`'s path globs,
+dev-environment state and the deployment-geometry parse the deploy cross-checks its flash
+address against.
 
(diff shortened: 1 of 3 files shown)


JojoEffect:
`/code-review high --fix` pass over this branch. Three findings, two fixed in `98f84d3`, one filed.

**Fixed — the missing-topic early return was described wrongly.** The test comment said that
without it "the payload comparison would run against a null entry and report a second failure".
It would not: this file runs under `Set-StrictMode -Version Latest`, where
`$Snapshot[$absent].Retained` throws `PropertyNotFoundException`. Checked, not assumed. So the
return is a crash guard, and calling it a tidier of messages is an invitation to fold it away —
which would take a device missing one attribute and turn it into a conformance run that dies with
no verdict. The reason now sits both in the test and next to the guard itself, and the case
asserts the failure's content rather than only that there is one of it.

**Fixed — the skill overstated what this covers.** It claimed `Get-AttributeFailure` was the first
piece of the conformance verdict covered at all. `Get-ConformanceCaptureSeconds` and the lifecycle
table already were, and the skill's own coverage list had simply omitted them. Narrowed to what is
true — the first *assertion* inside `Measure-HomieConformance` — and the list now names them.

**Filed, not fixed — #94.** `-AnyValue` and `-Expected` are mutually exclusive in meaning but not
in binding, so `-AnyValue -Expected 'ready'` is accepted and the payload comparison is silently
skipped. Latent: no call site does it. Not fixed here because two parameter sets is a *binding*
change, the 19 call sites are not exercised by the desk suite, and a set that binds wrongly at one
of them would be desk-green and only surface on a hardware run — which would have quietly
invalidated this PR's 5/5 verification for a hazard with no current caller. #94 says to fold it
into whichever change next takes this file to hardware.

Host-side suite after the fixes: **157 passed, 0 failed**. No production behaviour changed by
`98f84d3` — comments, one strengthened test assertion, and the skill.

Issue #95, opened 2026-09-03
Test-Setup.ps1 checks mosquitto.exe only, so a missing mosquitto_sub/mosquitto_pub passes the preflight and fails mid-run


JojoEffect:
`Test-Setup.ps1` checks exactly one mosquitto binary:

```powershell
$mosquitto = Join-Path $mosquittoDir 'mosquitto.exe'
if (Test-Path $mosquitto) {
    Add-Result -Name 'mosquitto.exe' -Status 'OK' -Detail $mosquitto
}
```

`scripts\Test-Setup.ps1:135-144`. Three binaries are actually required, and the other two are the ones nothing verifies:

| Binary | Needed by | Checked by the preflight |
|---|---|---|
| `mosquitto.exe` | `Start-DevEnv.ps1` (the broker) | yes |
| `mosquitto_sub.exe` | `Start-DevEnv.ps1` (the `homie/#` subscriber), `Run-IntegrationTests.ps1` (`Start-HomieCapture`, so every conformance snapshot) | **no** |
| `mosquitto_pub.exe` | `Run-IntegrationTests.ps1` (`Publish-HomieCommand`, so every `/set` the conformance check drives) | **no** |

So a Mosquitto installation carrying the broker but not the clients — a server-only package, a partial install, a `SMARTHOME_MOSQUITTO_DIR` pointing at a directory holding only `mosquitto.exe` — passes `Test-Setup.ps1` with `17 checked, 0 failed, 0 warned` and then fails at the first workflow that needs a client.

This is exactly the failure mode `Test-Setup.ps1` exists to remove. From `CLAUDE.md`: it "reports everything the other scripts assume exists on this machine ... all at once, rather than one abort at a time", and "It reports rather than aborts, on purpose: `Import-SmartHomeLocalEnv` exits on the first missing file, so a workflow that relies on it discovers one gap, stops, and hides the next." A preflight that passes on two thirds of a dependency hides the same gap it was written to surface.

The symptom is worse in the integration suite than in `Start-DevEnv.ps1`. `Start-DevEnv.ps1` names the missing binary at once. `Run-IntegrationTests.ps1` reaches `mosquitto_sub.exe` inside `Start-HomieCapture`, i.e. mid-run, after a build and a flash — and it reaches `mosquitto_pub.exe` later still, inside the conformance `/set` round trip, roughly 90 seconds into a `HomieClientCheck` run that has already deployed to the device.

## Evidence

[...]

Issue #71, opened 2026-08-31
Restore-Packages.ps1 and Test-Setup.ps1 glob with -Path, so a checkout path containing [ or ] finds zero packages.config


JojoEffect:
`Restore-Packages.ps1` and `Test-Setup.ps1` glob with `-Path`, so a checkout whose path contains `[` or `]` finds zero `packages.config`

## What is wrong

Both scripts locate the projects to work on with a wildcard-interpreting `-Path`:

- `scripts\Restore-Packages.ps1:47` — `$configs = Get-ChildItem -Path $repoRoot -Filter 'packages.config' -Recurse -File | ...`
- `scripts\Test-Setup.ps1:240` — `$configs = @(Get-ChildItem -Path $repoRoot -Filter 'packages.config' -Recurse -File -ErrorAction SilentlyContinue | ...)`

`-Path` treats `[` and `]` as wildcard character-class syntax. Both characters are legal in a Windows directory name, so a repository cloned to something like `C:\repos\SmartHome [wip]` makes the glob match nothing.

## Reproduction

Measured in a worktree created at `.claude\worktrees\zz-probe [x]`:

```
Get-ChildItem -Path        <root> -Filter packages.config -Recurse -File   ->  0
Get-ChildItem -LiteralPath <root> -Filter packages.config -Recurse -File   ->  14
```

The two scripts then fail differently, and neither points at the cause:

**`Restore-Packages.ps1` crashes.** `$configs` is `$null`, and line 50 does `$($configs.Count)`, which under `Set-StrictMode -Version Latest` throws:

```
Die Eigenschaft "Count" wurde für dieses Objekt nicht gefunden. Vergewissern Sie sich, dass die Eigenschaft vorhanden ist.
```

Nothing in that message mentions paths, wildcards, or `packages.config`. (The same line is fragile for an unrelated reason: in Windows PowerShell 5.1 a single match is a bare `FileInfo`, which also has no `.Count`. `Test-Setup.ps1` already wraps its equivalent in `@()`; `Restore-Packages.ps1` does not.)

**`Test-Setup.ps1` reports the opposite of the truth.** Its `@()` wrapper means `$configs.Count` is 0, so it takes the `$configs.Count -eq 0` branch and emits

```
[WARN] packages\    no packages.config found
      Unexpected -- check this is a full checkout
```

on a complete, fully restored checkout — a preflight passing (WARN does not fail the run) while silently skipping the check it exists to perform.

## What closing it would look like

[...]

PR #96, opened 2026-09-03
refactor(scripts): share the mosquitto lookup and the subscriber quoting

### LINKED ISSUE #85: Five duplications between Run-IntegrationTests.ps1 and Start-DevEnv.ps1 that #74 did not absorb

JojoEffect:
Items **1, 2 and 5** of #85 — the two cross-file duplications and the dead duplicate declaration. Items 3 and 4 (the polling helper behind the three waits, the retry helper behind the two rounds) are deliberately **not** here; they are the next two slices.

Does not close #85.

## What was duplicated

`Run-IntegrationTests.ps1` and `Start-DevEnv.ps1` each resolved a mosquitto binary under `SMARTHOME_MOSQUITTO_DIR` with a byte-identical guard and remediation message, and each rendered the subscriber's argument list for `cmd.exe` with the same two lines. Both files carried a comment admitting it — *"Same guard and remediation Start-DevEnv.ps1 gives the same two binaries"*, *"Quoting idiom matches Start-DevEnv.ps1's subscriber launch"* — and nothing reconciled them, so a change to the quoting rule or the guard in one file would have kept the old behaviour in the other, silently.

## What replaces it

Two functions in `Common.ps1`, beside `Get-SmartHomeSubscriberArguments`, which already owned the argument list the rendering flattens:

- **`Get-SmartHomeSubscriberArgumentString -Port`** — the arguments as one `cmd.exe` string, quoting only the values carrying whitespace, a slash or a `#`.
- **`Get-SmartHomeMosquittoTool -Name [-Directory]`** — the path, or the run stops with the same message. `-Directory` defaults from `SMARTHOME_MOSQUITTO_DIR`.

`-Directory` is the seam that makes the guard testable at all: without it every case would pass or fail on how Mosquitto happens to be installed on the machine running them. The new cases pass on a checkout with no `SMARTHOME_MOSQUITTO_DIR` set — which is CI's condition, and was this session's, confirmed by a mutation that removed the parameter and made three cases fail with `Missing environment variable: SMARTHOME_MOSQUITTO_DIR`.

`Test-Setup.ps1` deliberately does **not** read `Get-SmartHomeMosquittoTool`: it reports rather than aborts, by design, and reads values parsed out of `local.env.ps1` rather than the environment. Left alone.

## The declaration (#85 item 5), and the half of it that is not a defect

`$script:currentPhase = $null` did appear twice at file scope, both assigning `$null`. One is gone. The one kept is the earlier one — **not** the one #85 suggested — because its comment carries the reason the declaration has to exist before any conformance check does: `Save-SnapshotEvidence` reads it from `Stop-HomieCapture`'s `finally` on **every window of every test**, including runs that never enter the conformance check, and an unset variable there is a terminating error under `Set-StrictMode -Version Latest` — inside a `finally`, where it would replace the exception already in flight. The phase-timing block now points at it.

**`$script:conformancePhases` is not a duplicate and was left alone.** #85 says both pairs are at file scope; the second assignment is inside `Measure-HomieConformance` (confirmed with the AST, not by reading indentation), so it is a per-run reset. The file-scope one is load-bearing separately: `Write-ConformancePhaseBreakdown` runs from `Invoke-HomieConformanceCheck`'s `finally` and reads `.Count`, which is reached when the check fails *before* the measurement starts. Deleting either would have been a defect. Recorded on #85 as a correction to its own body.

## Verified

**Behaviour is unchanged, and that was measured rather than assumed.** The rendered subscriber arguments were compared against the old expression, evaluated side by side for two ports: byte-identical (`-ceq`) both times — `-h 127.0.0.1 -p 1883 -t "homie/#" -F "%t %r %p"`. The missing-tool path still ends the run with exit code 1 and the same message.

- **`.\scripts\Run-ScriptTests.ps1` — 168 passed, 0 failed**, ~9s, no device. Up from 157: 4 cases for the rendering, 7 for the tool lookup (5, plus 2 regression cases from the review pass below).
- **Three mutations, all caught**, because a case that cannot fail proves nothing:
  - narrowing the quoting rule from `[\s/#]` to `[\s]` → 2 failures, naming `-t homie/#` unquoted;
  - deleting the `-Directory` parameter → 3 failures;
  - replacing the not-found guard with `if ($false)` → 2 failures, including the child-process case reporting `expected 1, got 0`.
- **The dev environment, started and stopped for real** (no device): broker up, two publishes, both back out of the `homie/#` subscriber's log in the `%t %r %p` layout — which is the wildcard subscription surviving the quoting end to end, not just the string matching an expectation.
- **Both rewired call sites inside `Run-IntegrationTests.ps1`, exercised at desk against a live broker and no device**, by dot-sourcing the script through its own guard: `Get-HomieRetainedSnapshot` returned 2 entries with the retain flag set, and `Publish-HomieCommand` produced a command seen live inside a window that had waited for its subscriber. Two evidence files were written, so `Save-SnapshotEvidence` read `$script:currentPhase` with only the one declaration left.
- **Hardware: the full suite, all 5 PASS — twice.** Once on `0cd59f8` (201s) and again on `7483a28` after the review fixes (**197s** — WifiCheck 26s, MqttCheck 24s, Bmp280Check 24s, HomieClientCheck 70s, MqttReconnectCheck 53s), so the result stands for the diff as it is now rather than for an earlier version of it. `HomieClientCheck` is the one that matters here: 53 topics in the retained store both times, every phase normal, and the re-announce after the broker was replaced. Every deploy took the #46 erase path with no fallback and no warnings, and the new mosquitto-client pre-flight passed silently ahead of the first build.

The suite leaves `MqttReconnectCheck` flashed, as it always does. RoomSensor has not been redeployed.

## Filed

[#95](https://github.com/JojoEffect/SmartHome/issues/95) — `Test-Setup.ps1` checks `mosquitto.exe` only, so a Mosquitto install missing `mosquitto_sub.exe` or `mosquitto_pub.exe` passes the preflight and then fails inside `Start-DevEnv.ps1` or a conformance snapshot, which is the exact failure the preflight exists to pre-empt. Noticed while giving those two binaries a shared lookup; verified against the source, not fixed here.

### DIFF
--- a/scripts/Common.ps1
+++ b/scripts/Common.ps1
@@ -682,6 +682,55 @@ function Get-SmartHomeSubscriberArguments {
     return @('-h', $SmartHomeLocalBrokerHost, '-p', $Port, '-t', $SmartHomeHomieTopic, '-F', '%t %r %p')
 }
 
+function Get-SmartHomeSubscriberArgumentString {
+    param(
+        [Parameter(Mandatory = $true)]
+        [string]$Port
+    )
+
+    # The arguments above, rendered as one string for cmd.exe. Both scripts that start a
+    # subscriber redirect its output, and both do the redirect in cmd.exe rather than with
+    # Start-Process -RedirectStandardOutput (see Start-DevEnv.ps1's header for why), so
+    # both need the argument list flattened.
+    #
+    # Quoting is the part worth sharing. Only the values carrying whitespace, a slash or a
+    # '#' need it -- 'homie/#' is the one that does today -- and the rule lived character
+    # for character in both files, in the half that breaks first if the argument list ever
+    # grows a value needing different quoting.
+    return (Get-SmartHomeSubscriberArguments -Port $Port |
+        ForEach-Object { if ($_ -match '[\s/#]') { '"{0}"' -f $_ } else { $_ } }) -join ' '
+}
+
+function Get-SmartHomeMosquittoTool {
+    param(
+        [Parameter(Mandatory = $true)]
+        [string]$Name,
+
+        # Defaulted from SMARTHOME_MOSQUITTO_DIR rather than read unconditionally, so a
+        # caller -- a test, above all -- can ask about a directory without the machine's
+        # own configuration deciding the answer.
+        [string]$Directory
+    )
+
+    if (-not $PSBoundParameters.ContainsKey('Directory')) {
+        $Directory = Get-RequiredEnvValue -Name 'SMARTHOME_MOSQUITTO_DIR'
+    }
+
+    $path = Join-Path $Directory $Name
+
+    # Without this guard a wrong SMARTHOME_MOSQUITTO_DIR surfaced as a raw
+    # CommandNotFoundException -- or, for Run-IntegrationTests.ps1's retained snapshot, as
+    # an empty read reported as a conformance FAIL, i.e. a machine-config problem blamed on
+    # the device. exit rather than throw, matching Get-RequiredEnvValue above: this is a
+    # setup fault, and every caller's answer to it is to stop.
+    if (-not (Test-Path $path)) {
+        Write-Error ("Not found: {0}`nCheck SMARTHOME_MOSQUITTO_DIR in local.env.ps1." -f $path)
+        exit 1
+    }
+
+    return $path
+}
+
 # ── Dev environment: state ────────────────────────────────────────────────────
 # Start-DevEnv.ps1 records what it spawned so Stop-DevEnv.ps1 (or the
 # integration-test runner) can shut it down from a different shell. Keyed by
--- a/scripts/Common.ps1
+++ b/scripts/Common.ps1
@@ -708,11 +708,17 @@ function Get-SmartHomeMosquittoTool {
 
         # Defaulted from SMARTHOME_MOSQUITTO_DIR rather than read unconditionally, so a
         # caller -- a test, above all -- can ask about a directory without the machine's
-        # own configuration deciding the answer.
+        # own configuration deciding the answer. Empty and unsupplied mean the same thing
+        # here; see below.
         [string]$Directory
     )
 
-    if (-not $PSBoundParameters.ContainsKey('Directory')) {
+    # Tested by value rather than with $PSBoundParameters.ContainsKey, because a caller
+    # reading the directory out of a config file and passing it straight through binds the
+    # parameter with an empty string. That satisfies ContainsKey, skips the fallback, and
+    # dies in Join-Path's parameter binding -- a raw ParameterBindingValidationException
+    # naming Join-Path, which is the kind of message the guard below exists to replace.
+    if ([string]::IsNullOrWhiteSpace($Directory)) {
         $Directory = Get-RequiredEnvValue -Name 'SMARTHOME_MOSQUITTO_DIR'
     }
 
@@ -723,7 +729,12 @@ function Get-SmartHomeMosquittoTool {
     # an empty read reported as a conformance FAIL, i.e. a machine-config problem blamed on
     # the device. exit rather than throw, matching Get-RequiredEnvValue above: this is a
     # setup fault, and every caller's answer to it is to stop.
-    if (-not (Test-Path $path)) {
+    #
+    # -LiteralPath, not -Path: an install directory carrying [ or ] is a wildcard pattern
+    # to Test-Path, so a mosquitto.exe that demonstrably exists reads as missing and the
+    # remediation then points at a path that is already correct (issue #71's defect class,
+    # which is why Set-TestFileContent and Import-SmartHomeLocalEnv are -LiteralPath too).
+    if (-not (Test-Path -LiteralPath $path)) {
         Write-Error ("Not found: {0}`nCheck SMARTHOME_MOSQUITTO_DIR in local.env.ps1." -f $path)
         exit 1
     }
--- a/scripts/Run-IntegrationTests.ps1
+++ b/scripts/Run-IntegrationTests.ps1
@@ -857,24 +857,6 @@ function Wait-ForAnnounceWitnessed {
     return $false
 }
 
-function Get-MosquittoTool {
-    param([string]$Name)
-
-    $dir = Get-RequiredEnvValue -Name 'SMARTHOME_MOSQUITTO_DIR'
-    $path = Join-Path $dir $Name
-
-    # Same guard and remediation Start-DevEnv.ps1 gives the same two binaries. Without
-    # it a wrong SMARTHOME_MOSQUITTO_DIR surfaced as a raw CommandNotFoundException --
-    # or, for the retained snapshot, as an empty read reported as a conformance FAIL,
-    # i.e. a machine-config problem blamed on the device.
-    if (-not (Test-Path $path)) {
-        Write-Error ("Not found: {0}`nCheck SMARTHOME_MOSQUITTO_DIR in local.env.ps1." -f $path)
-        exit 1
-    }
-
-    return $path
-}
-
 # How long a snapshot subscriber listens. Three seconds, and it stays three seconds --
 # a fixed value rather than a parameter, because there is nothing here for a caller to
 # choose. Two attempts at shortening it both broke the conformance check, and the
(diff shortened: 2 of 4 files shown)


JojoEffect:
## Review pass (`/code-review xhigh --fix`), applied in `7483a28`

Eight findings; seven fixed, one skipped. Two were real defects in the shipped code, both reproduced against the running function before being fixed, and both now mutation-checked.

**1. The shared lookup carried in #71's defect class.** `Get-SmartHomeMosquittoTool` tested the resolved path with `Test-Path -Path`, so a Mosquitto directory containing `[` or `]` was read as a wildcard pattern. Reproduced: created `...\mosquitto[2]\mosquitto.exe`, confirmed `Test-Path -LiteralPath` sees it, then watched the function report `Not found: ...\mosquitto[2]\mosquitto.exe` — i.e. the guard blaming `SMARTHOME_MOSQUITTO_DIR` for a path that is already correct, which is exactly the misleading remediation it exists to prevent. Now `-LiteralPath`, pinned by a case with a bracketed fixture directory. This came in with the moved line; `origin/main` has it in both scripts.

**2. The environment fallback tested the binding, not the value.** `$PSBoundParameters.ContainsKey('Directory')` is satisfied by `-Directory ''`, so the fallback was skipped and `Join-Path` failed its own parameter binding — `Das Argument kann nicht an den Parameter "Path" gebunden werden...`, with no mention of `SMARTHOME_MOSQUITTO_DIR`. Confirmed for both `''` and `$null`. Now `[string]::IsNullOrWhiteSpace($Directory)`, which also drops the `$PSBoundParameters` indirection. That is the shape `Test-Setup.ps1` would use if it adopted the helper (#95).

**3. The pre-flight now resolves both mosquitto clients.** Same principle the pre-flight already states for the marker names — a desk operation, worth failing on before the first 90-second flash. Nothing else caught them that early: `Start-DevEnv.ps1` never resolves `mosquitto_pub.exe` at all, and with `-NoBroker` it does not run, so the first report used to come from `Publish-HomieCommand` inside `HomieClientCheck`'s `/set` phase. Placement checked by line number: the new call sits at 2328, ahead of the debug-monitor build (2334), the broker (2351) and the first flash (2385).

**Three in the new cases themselves.** The child-process probe interpolated paths into single-quoted literals without doubling apostrophes — a temp path under `C:\Users\O'Brien` would have produced a probe that fails to parse, and a parse error *also* exits 1, so the case would have passed while proving nothing. The quoting case copied the function's own `[\s/#]` regex to build its expectation, reintroducing inside the test suite the exact unreconciled duplication this PR removes; it now states the rule as a character set, so the two formulations disagree loudly instead of drifting together. And the stderr read is now as guarded as the stdout read beside it.

**One convention note, fixed with a comment.** The probe uses `Start-Process -RedirectStandardOutput/-RedirectStandardError`, the idiom CLAUDE.md:74-79 tells `Start-DevEnv.ps1` never to use. It is safe here because `-Wait` means no child outlives the call — verified by running the suite with its output piped — but the safety condition was unwritten, and the next person to copy the case into something that backgrounds inherits the hang. It is written down now.

**One skipped:** `Test-Setup.ps1` still hand-builds its own `Join-Path $mosquittoDir 'mosquitto.exe'`. It cannot call the shared helper as written — that helper exits, and this script's whole contract is to report every gap and keep going. Left for [#95](https://github.com/JojoEffect/SmartHome/issues/95), which now [records](https://github.com/JojoEffect/SmartHome/issues/95#issuecomment-5524801864) that fix 1 turned that divergence from latent into real: on a bracketed install the scripts resolve the binary and the preflight reports FAIL.

## Re-verified

- `.\scripts\Run-ScriptTests.ps1` — **168 passed, 0 failed** (was 166; two regression cases).
- **Both new cases mutation-checked**: reverting `-LiteralPath` fails *finds a tool under a directory whose name contains [ or ]*; reverting the value test fails *treats an empty directory as no directory rather than as a path* with the `ParameterBindingValidationException` above.
- The live-broker desk check re-run: 2 retained snapshot entries, the `/set` seen live in a waited window, 2 evidence files.

**The hardware run in the description predates these fixes.** Of what changed since, only the pre-flight addition is on a runtime path a suite run would newly execute, and it is two calls to a function whose failure path is covered in a child process; the two `Common.ps1` fixes are no-ops on a bracket-free directory with no caller passing `-Directory`. Saying so rather than letting the earlier result stand for the current diff.

JojoEffect:
## Hardware re-run on `7483a28`

The gap the review comment above flagged is closed: the full suite ran again on the fixed code, not just on `0cd59f8`.

**All 5 PASS, 197s, exit 0** (previous run on `0cd59f8`: 201s).

```
  WifiCheck            PASS   26s (21s deploy)
  MqttCheck            PASS   24s (19s deploy)
  Bmp280Check          PASS   24s (19s deploy)
  HomieClientCheck     PASS   70s (21s deploy)
  MqttReconnectCheck   PASS   53s (21s deploy)
  TOTAL                       197s (101s deploy)
```

The parts worth naming, because they are what the fixes touched:

- **The new mosquitto-client pre-flight ran and passed silently**, ahead of the debug-monitor build and 57 lines before the first flash. Nothing in the output between the broker start and the first build, which is the correct appearance of a check that finds both binaries.
- **`HomieClientCheck`** — 53 retained topics, same as before; announce witnessed live at `$state=init`; phase breakdown normal (announce 9.8s / 2 snapshots, /set 3.7s, out-of-format 3.8s, lifecycle 17.5s / 5, re-announce 13.6s / 2). That exercises `Get-SmartHomeMosquittoTool` for both clients and `Get-SmartHomeSubscriberArgumentString` on every snapshot.
- **`MqttReconnectCheck`** — heartbeat 1 → 3 → 5 across 3s and 20s outages, subscription replay confirmed with `echo-5`.
- **Every deploy took the #46 erase path**, no fallback, no warnings — e.g. `Deployment area past 182944 bytes is erased; flashing the image as built.`

Desk verification was re-run alongside it: `Run-ScriptTests.ps1` **168 passed, 0 failed**, and the live-broker check (2 retained snapshot entries, the `/set` seen live in a waited window, 2 evidence files). The description's verification section now reflects both hardware runs.

Issue #105, opened 2026-09-04
Home Assistant integration: HA does not speak Homie, but a first-party nanoFramework HA-discovery binding already exists


JojoEffect:
Research finding from a session asking how to integrate a nanoFramework device into Home Assistant over MQTT. Filing per CLAUDE.md's "File what you find" — no code changed, and this would otherwise only exist in a session summary.

## What was verified

**1. Home Assistant has no native Homie support.** RoomSensor's `homie/room-sensor-office/#` tree is invisible to HA as it stands. HA's own self-description spec is **MQTT Discovery**, which is the direct analogue to Homie but a different wire format. Only community bridges (`labodj/homie-home-assistant-discovery`, `elbowz/ha-homie`) map Homie → HA; neither is a HA core integration.

**2. The HA MQTT Discovery spec has two forms**, verified against `homeassistant/components/mqtt/schemas.py` on HA core `dev` (not just the prose docs):

- Single-component: `homeassistant/<component>/[<node_id>/]<object_id>/config`. `device` is *optional*; when present it needs at least one of `identifiers`/`connections` (`validate_device_has_at_least_one_identifier`). `name` inside the device block is `vol.Optional` — so the common "one entity carries the full device block, the rest link by `ids` only" pattern is valid, despite docs prose suggesting `name` became mandatory.
- Device-based: `homeassistant/device/<object_id>/config`, added in HA **2024.11**. `DEVICE_DISCOVERY_SCHEMA` makes `device`, `components` and `origin` all **required**, each component config requires `platform` (`p`), and entity platforms require `unique_id`. `MQTT_ORIGIN_INFO_SCHEMA` requires `name`.

Discovery payloads should be published **retained**, or (re)published on HA's birth message at `homeassistant/status` (`online`). Availability is an LWT on a topic named by `availability_topic`.

**3. There is already a first-party nanoFramework binding**: `nanoFramework.Iot.Device.HomeAssistant`, source at `nanoFramework.IoT.Device/devices/HomeAssistant`, published on NuGet (latest **1.0.30**; also 1.0.1 / 1.0.20 / 1.0.26). ~3.2k lines. It implements single-component discovery, retained discovery payloads, LWT availability, `homeassistant/status` restart detection, a device-id-derived stable MQTT client id, and entity types Switch / Number / Sensor / BinarySensor / Button / Select / Text / Light / Cover / Climate, plus sensor presets for temperature, humidity and pressure — exactly RoomSensor's three properties.

[...]

PR #106, opened 2026-09-04
feat: announce Homie devices to Home Assistant on the same MQTT session

JojoEffect:
## Summary

Adds `SmartHome.HomeAssistant`, which publishes a Homie device to Home Assistant's MQTT Discovery convention over the connection `HomieClient` already opened — no second session, no duplicated state, no new traffic beyond one retained config per property.

Prototypes path 1 of #105.

Part of #105

## Changes

- **`src/common/HomeAssistant/`** — new library. Dependency extends the existing one-way chain: `SmartHome.HomeAssistant` → `SmartHome.Homie` → `SmartHome.Mqtt`. Nothing references back and the Homie library does not know this one exists.
  - `DiscoveryMapper` — pure function from a Homie `Device` to discovery messages. Reads `$datatype`, `$unit`, `$format` and `$settable` off each property and emits one retained config whose `state_topic` is that property's own value topic and whose `command_topic` is the `/set` topic `HomieClient` already subscribes to. **The device publishes no second copy of any value**, and a property added to the Homie model gets an entity with no edit anywhere else.
  - `HomeAssistantAnnouncer` — borrows an already-connected `IReconnectingMqttClient`; publishes the configs and keeps them alive. Deliberately not a client: MQTT allows one will per connection and Homie has already spent it on `$state` = `lost`.
  - `JsonWriter` — ~150 lines, hand-rolled rather than adding `nanoFramework.Json` as a further assembly for a few dozen members of flat generated JSON. Same call the upstream nanoFramework binding made.
  - `DeviceClass`, `Component`, `EntityOverride`, `HomeAssistantSettings`, `HomeAssistantTopics`.
- **`src/devices/RoomSensor/`** — dual-announces. Temperature, humidity and pressure reach Home Assistant as sensors with the correct device classes, an `expire_after` derived from the measurement interval, and availability that follows the Homie will it already had. The announce is wrapped in a `try`/`catch`: the device's job is publishing readings, and Main's catch would otherwise turn a Home Assistant problem into a reboot.
- **`src/tests/Unit/`** — 31 tests over the mapping and both re-announce paths.
- **`.github/workflows/pr-title.yml`** — adds `homeassistant` to the allowed scope list, since this branch adds that area under `src/common` and the list tracks the layout. It applies to future pull requests, not this one: the check runs on `pull_request_target`, so the workflow is read from the base branch.
- **`CLAUDE.md`** — layout entry plus a section on the two decisions that carry the design.

### Two design points worth reviewing

**Availability comes from `$state`, not a second will.** `DiscoveryMapper.AvailabilityTemplate` maps the six Homie states onto Home Assistant's online/offline via an `availability_template`. `ready`, `alert` and `sleeping` are available — the device is connected and answering in all three; `init`, `disconnected` and `lost` are not. `lost` is the Homie last will, so a device that drops off the network goes unavailable in Home Assistant with no extra machinery. This is the claim that lets one session carry both conventions.

**Single-component discovery**, not the device-based form Home Assistant 2024.11 added. Device discovery is one message instead of N, but it builds the whole device's configuration in memory at once; N small messages built and released in turn is the cheaper shape on this hardware, and it still works on installations older than 2024.11.

### Review round (`82f740e`)

A max-effort review found seven defects, all the same shape: the mapper read a Homie declaration differently from the code that enforces it, so Home Assistant was offered something the device would then refuse — or nothing at all. All seven are fixed on `82f740e`, with five new tests.

- **Enum options were not trimmed**, and the comment saying so asserted the *reverse* of what `EnumProperty.Validate` does — it trims each declared value before comparing. A `$format` of `"eco, sport"` (a spelling that file explicitly supports) offered `" sport"`, and selecting it came back as a payload the property rejected. The control was dead for that value.
- **The range was read one end at a time**, untrimmed, unpaired and shape-unchecked, diverging from `PropertyBase.TryGetDeclaredRange` in three directions: `"5 : 30"` is enforced by the device and was dropped here; `"30:5"` is *not* enforced by the device and was emitted as `min:30,max:5`, which HA's number schema rejects outright; and `".5"`, `"1."`, `"1.2.3"` passed a digit-counting check and went out as raw JSON tokens, producing a payload HA cannot parse at all.
- **`stat_cla: measurement` went out for every numeric sensor**, including the cumulative classes `DEVICE_CLASS_STATE_CLASSES` forbids it for — reachable from a litre/gallon unit via `volume`, and from `SetDeviceClass` for `energy`, `water`, `gas`. The unit table was checked when this shipped; the state-class table was not.
- **A settable enum with no `$format`** became a `select` with no `options`, which the select schema rejects, so the entity never appeared. It now maps to `text`, which is what `EnumProperty` accepts in that case.
- **`Announce()` was documented as never throwing** but only guarded its publishes, leaving the per-property JSON allocation free to escape onto M2Mqtt's dispatch thread — which treats that as a dead connection.
- **`Detach()` guarded handler removal on a flag `Attach()` sets only after its SUBSCRIBE**, so a throwing SUBSCRIBE left both handlers live and made `Detach()` a silent no-op.

One further finding was **not** fixed here and is filed as #107: `object_id` joins three hyphen-legal Homie ids with `-`, so `tank-level`/`litres` and `tank`/`level-litres` collide onto one config topic. The fix changes every discovery topic and `unique_id`, which orphans already-published retained configs — a migration decision rather than a review fix.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / chore
- [ ] Infrastructure / tooling

## Testing Done

**Nothing has been run on hardware.** That remains the gap, and it is the only one: whether Home Assistant actually accepts these configs and renders the entities, whether the availability template behaves across a real Homie lifecycle, and whether the re-announce fires when a real broker or a real Home Assistant restarts — none of that is provable without a device and an HA instance.

**CI on `82f740e`: 92 of 92 unit tests pass** on the nanoclr virtual device, and the solution builds. That is the 87 from the previous round plus the five new cases the review round added.

(For the record on the way here: the first run, on `0042cd4`, failed two tests — arithmetic in the test rather than behaviour in the mapper, since the fixture builds seven properties and the assertions said six and five. `cf054c0` fixed that by counting off the device, so the expectation states the claim and cannot drift, and ran 87/87.)

Both rounds were written **without a C# toolchain** — there is no `dotnet`/`csc`/`mono` in the environment — so CI has been the first compile check each time. What was verified locally instead:

- **The payloads were checked as data**, before and after the review round. The pure logic (`JsonWriter`, `IsJsonNumber`, `TryGetRange`, `OptionsOf`, the payload construction) was ported line-for-line to a scratch script and run against RoomSensor's actual device model and every component shape the tests assert; each result was parsed with a real JSON parser. All parse. `IsJsonNumber` was table-tested against `.5`, `1.`, `1.2.3`, `-.5`, `1e3` and a trailing space, all correctly rejected.
- **Every emitted key was checked against Home Assistant's abbreviation table** (`uniq_id`, `stat_t`, `avty_t`, `avty_tpl`, `cmd_t`, `unit_of_meas`, `dev_cla`, `stat_cla`, `exp_aft`, `pl_on`/`pl_off`, `stat_on`/`stat_off`, `dev`, `o`, and the `ids`/`mf`/`mdl`/`sw` inside them). All resolve.
- **Both Home Assistant tables were read from `home-assistant/core` source** rather than recalled: `DEVICE_CLASS_UNITS` for the unit inference, and `DEVICE_CLASS_STATE_CLASSES` for the state-class fix (`VOLUME`, `ENERGY`, `WATER`, `GAS` are `{TOTAL, TOTAL_INCREASING}`; everything else this library produces is `{MEASUREMENT}`).
- **The two Homie-side claims behind the range and enum fixes were read from this repo's own source**: `EnumProperty.Validate` trims each `$format` candidate, and `PropertyBase.TryGetDeclaredRange` trims both ends and requires `min <= max`.

## Checklist

- [x] Relevant tests pass — 92/92 on `82f740e`
- [x] PR title follows Conventional Commits (it becomes the merge commit message)

No NuGet package versions changed: the new project pins byte-for-byte the same set `SmartHome.Homie` already does.

https://claude.ai/code/session_01AJDzPanF4sEb5ovHXeBpdX


JojoEffect:
Closing unmerged — the design is wrong for what this repository wants.

This ran Homie and Home Assistant **simultaneously**, with the Home Assistant side parasitic on Homie's topics: its `state_topic` was the Homie property topic, its `command_topic` the Homie `/set` topic, and its availability a template over `homie/<dev>/$state`. What is wanted instead is exactly **one** convention, chosen by configuration, with the device model shared and the protocol behind an injected adapter — so that Homie v4, Homie v5 and Home Assistant differ only in which adapter is constructed.

That is now #108, with sub-issues #109 → #112.

Nothing here is thrown away. The branch `claude/nanoframework-homeassistant-mqtt-x9j213` stays, and #111 lists what the Home Assistant adapter should lift from it rather than rediscover:

- the Homie→HA mapping (datatype + settable → component, `$format` → `min`/`max`/`step`/`options`, unit → device class);
- `DEVICE_CLASS_UNITS` and `DEVICE_CLASS_STATE_CLASSES`, both read from `home-assistant/core` source rather than recalled — the second is easy to miss, and is why `state_class: measurement` must not go out for `volume`, `energy`, `water` or `gas`;
- every discovery key checked against Home Assistant's own abbreviation table;
- `JsonWriter`, so `nanoFramework.Json` need not ship for a few dozen members of flat generated JSON;
- the seven defects a max-effort review found on this branch, each a trap the rewrite can walk into again. The most important is structural: a raw `string Format` re-parsed by each consumer diverged from the property that enforces it, three different ways. #109 makes formats structured in the model specifically to remove that whole class.

#107, the `object_id` collision, moves from "migration decision" to "fix it before anything is published" — #111 folds it in.

For the record, the state being closed: CI green on `82f740e`, 92/92 unit tests, mergeable, never run on hardware.

---
_Generated by [Claude Code](https://claude.ai/code)_

